PR #31, opened 2020-08-07
Enable strict typing of CSS variables through an enum

bradynpoulsen:


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@amino-ui/core",
   "version": "0.29.0",
   "description": "Core UI components for Amino",
-  "main": "index.js",
+  "main": "dist/index.js",
```
bradynpoulsen, line 5: This will enable imports to use `@amino-ui/core` instead of `@amino-ui/core/dist`. This should also improve the auto-completion generated by IDEs that do static analysis based on a package's declarations

PR #136, opened 2021-06-24
fix(layout): incr content bottom padding

anchung333:
### Description

This PR is related to a chore [insert dashboard-web PR here] as well as the below comment found in the ```Wrapper component``` in dashboard-web: 

image

This PR applies increased padding to the bottom of the ```Content``` component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "2.127.0",
+  "version": "2.128.0",
   "description": "Core UI components for Amino",
   "main": "dist/index.js",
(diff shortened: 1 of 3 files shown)


anchung333:
I might add a fix to this version bump that involves small screen and sidebar scrolling unless that is not desirable.

anchung333:
This PR will also address a bug with small screens brought up by @joshbeitler thanks for pointing it out. 

https://user-images.githubusercontent.com/57578761/123343658-6a534580-d50f-11eb-850a-cd2d0955e9b3.mov

PR #140, opened 2021-07-08
fix(slider): change position of indicator

anchung333:
### Jira

N/A

### Description

This PR changes the position of the the Indicator to relative, as setting it to absolute was causing some height calculation problems in Dashboard and as I want to avoid hard-coded heights, changed the Indicator position to relative to place the Indicator back into the document flow. 

### Before
image

Setting the position to relative should place the Indicator back into the document flow and have it be factored into height calculations and avoid unnecessary overflow for this and other cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "2.133.0",
+  "version": "2.134.0",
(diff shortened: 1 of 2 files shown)


joshbeitler:
Nice!

joshbeitler (approved):

PR #159, opened 2021-09-09
Build/ts eslints errors

anchung333:
### Description

This PR implements changes to provide a better experience for the developer; these changes include getting rid of prettier and eslint warnings and errors, as well as updating packages and identifying them as the correct type of dependency.


anchung333:
There are 3 eslint-warnings left that I am having some trouble with; feedback would be appreciated. thanks.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -88,7 +88,7 @@ export const Checkbox = ({
     multiline={!!subtitle}
     onClick={() => onChange(!checked)}
   >
-    <AminoCheckbox checked={checked}>
+    <AminoCheckbox checked={checked} id="amino-checkbox">
```
goodpickle, line 91: An id needs to be [globally unique](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id). What do you think about dynamically creating an id & `htmlFor`?

src/components/Radio/RadioGroup.tsx
```diff
@@ -32,6 +32,15 @@ export const RadioGroup = <T extends { label?: string; value?: string }>({
 
   const activeItem = items.find((_, i) => i === active);
 
+  const initial = items.findIndex(el => {
+    const value = itemValuePath ? el[itemValuePath] : el.value;
```
goodpickle, line 36: What do you think about cleaning this up similar to the `Select` component by removing itemValuePath & itemLabelPath?

goodpickle (changes requested):

src/components/Checkbox/Checkbox.tsx
anchung333: brain fart; you're right. good catch thanks!

src/components/Radio/RadioGroup.tsx
anchung333: done

goodpickle (approved):

PR #169, opened 2021-09-27
fix(radio): duplicate onchange call

goodpickle:
## Description

This PR fixes a duplicate on change call from the radio component.


ddinehart:
Why did you create a PR in the other one and link it to this instead of the other way around?

PR #172, opened 2021-11-04
Fix/layout nav height issue

anchung333:
## Jira issue
N/A

## Description
This PR should address the below issue in Dashboard when Dashboard pulls in the latest version of @amino-ui that incorporates the changes in this PR. I applied the below changes directly dashboard-web within the Browser Developer Tools which you can see in the `After` screenshot: 

Before:
image

After:
image

This PR also adds a floating `UserMenu` component to the Storybook that the `Layout` component can use in its footer to allow for a better testing experience.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "2.159.0",
+  "version": "2.160.0",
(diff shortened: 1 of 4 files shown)


ddinehart (approved):
This looks good Andy, great work.

PR #173, opened 2022-01-12
Fix/toasts wrapper z index

anchung333:
## Jira issue
[Homemade Merch/Overcast Stores: Chatty Broads 5101 & Space Gal 4023 : Cant create labels](https://myzonos.atlassian.net/browse/WS-341)

## Description
This PR adds a z-index to the toasts-wrapper div so it can now 'compete' with the `CoverSheet` component. The `Toast` component itself has a very high z-index, but due to [the stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), its parent div (`toasts-wrapper` class), which has a z-index of 0/unspecified, is hidden behind the `CoverSheet`, which has a z-index of 990 and is on the same stacking layer (refer to the stacking context link). This PR elevates the toasts-wrapper div above the `CoverSheet` so notifications/toasts will appear over `CoverSheet` instances.

### DIFF
--- a/src/styles/amino.css
+++ b/src/styles/amino.css
@@ -1,4 +1,4 @@
-@import url("https://rsms.me/inter/inter.css");
+@import url('https://rsms.me/inter/inter.css');
 
 html {
   font-size: var(--amino-type-scale-base);
@@ -110,6 +110,7 @@ h6 {
 
 .toasts-wrapper {
   position: fixed;
+  z-index: 9999999;
   bottom: var(--amino-space-double);
   right: var(--amino-space-double);
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "2.160.0",
+  "version": "2.161.0",
   "description": "Core UI components for Amino",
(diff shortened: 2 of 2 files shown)


src/styles/amino.css
```diff
@@ -110,6 +110,7 @@ h6 {
 
 .toasts-wrapper {
   position: fixed;
+  z-index: 9999999;
```
goodpickle, line 113: What is the Z-index you are building over?
9999999 seems excessive.

goodpickle (approved):
I will merge it, but I want you to revise the chosen z-index.

src/styles/amino.css
anchung333: > `z-index: 990;`

Is the z-index set for `CoverSheet.` 

> > z-index: 999999;

is the z-index set for the `Toast` itself. 

I do agree that it is excessive. I will look into standardizing z-indices in a future update.

anchung333: I do like clear enumerations like [this suggestion](https://stackoverflow.com/a/13591180)

PR #214, opened 2022-04-05
Feat/multi select

goodpickle:
## Jira issue

[Amino 3 - Select](https://myzonos.atlassian.net/browse/BBT-514)
[Amino 3 - Multi select](https://myzonos.atlassian.net/browse/BBT-516)

## Description

This PR adds the Select and Multi-Select to V3.


src/stories/ScratchPad.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <SecondOption label="Benin" />
+          <SecondOption label="Austria" />
+          <SecondOption label="Greenland" />
+          <SecondOption label="Iceland" />
+          <SecondOption label="United States" />
+        </Container>
+      </StyledWrapper>
+    </VStack>
+  );
+};
```
Thienhuynh95, line 112: I noticed that this ScratchPad file is just exporting the static HTML block, with no `props` available. Will we use this in the future?

goodpickle: Thanks for the suggestion of moving it away from the other Stories. I think it would have been confusing having it in line with the rest.

Thienhuynh95 (approved):
Really good stuff!!!! Well done!!!!

PR #217, opened 2022-04-06
Feat/rich radio customization

Thienhuynh95:
## Description

This PR adds prop to customize rich radio options

### DIFF
--- a/src/components/RichRadio/RichRadio.tsx
+++ b/src/components/RichRadio/RichRadio.tsx
@@ -91,7 +91,7 @@ type RichRadioItemType = {
 
 export type RichRadioProps = {
   onChange: (newVal: string) => void;
-  renderTextWrapper?: (option: RichRadioItemType) => ReactNode;
+  renderCustomText?: (option: RichRadioItemType) => ReactNode;
   items: RichRadioItemType[];
   value: string;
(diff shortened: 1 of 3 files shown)


src/components/RichRadio/RichRadio.tsx
```diff
@@ -91,28 +91,36 @@ type RichRadioItemType = {
...
       <VStack spacing="space-half">
         {items.map(item => (
           <StyledItem value={item.value} key={item.value}>
-            <div>
-              <Label>{item.label}</Label>
-              {item.subtitle && <Subtitle>{item.subtitle}</Subtitle>}
-            </div>
+            {renderTextWrapper ? (
+              renderTextWrapper(item)
+            ) : (
```
goodpickle, line 118: What do you think of `renderCustomLabel` or `renderCustomText`?
We really are doing more than just rendering a wrapper. 🤔

goodpickle (changes requested):

src/components/RichRadio/RichRadio.tsx
Thienhuynh95: That's true! Thanks 😉

goodpickle (approved):

PR #221, opened 2022-04-11
feat(rich-radio):optional tooltip for radio option

Thienhuynh95:
## Description

This PR adds optional tool tip and tooltip setting to "items" property and bumps version to 3.0.19

### DIFF
--- a/src/components/RichRadio/RichRadio.tsx
+++ b/src/components/RichRadio/RichRadio.tsx
@@ -1,4 +1,5 @@
 import React, { ReactNode } from 'react';
+import ReactTooltip, { TooltipProps } from 'react-tooltip';
 
 import { Indicator, Item, Root } from '@radix-ui/react-radio-group';
 import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


src/components/RichRadio/RichRadio.tsx
```diff
@@ -112,7 +115,14 @@ export const RichRadio = ({
...
       <VStack spacing="space-half">
         {items.map(item => (
-          <StyledItem value={item.value} key={item.value}>
+          <StyledItem
+            value={item.value}
+            data-tip={item.tooltip}
+            key={item.value}
+          >
+            {item.tooltip && (
+              <ReactTooltip {...{ effect: 'solid', ...item.tooltipSetting }} />
```
goodpickle, line 124: What do you think about using this syntax for prop spreading?
```suggestion
              <ReactTooltip effect="solid" {...item.tooltipSetting } />
```

goodpickle (changes requested):

src/components/RichRadio/RichRadio.tsx
Thienhuynh95: That definitely looks better than mine 😄 Thank you

goodpickle (approved):

PR #233, opened 2022-04-15
Feat/add type prop to button

Thienhuynh95:
## Description

This PR adds type prop to Button component


src/components/Button/Button.tsx
```diff
@@ -224,6 +225,7 @@ export const Button = ({
   size,
   tabIndex,
   tooltip,
+  type,
```
goodpickle, line 228: By setting the default here, you should be able to avoid the `type || 'button'` throughout the file.
```suggestion
  type = 'button',
```

goodpickle (changes requested):

PR #270, opened 2022-05-27
v3.0.50

Thienhuynh95:
## Description

This PR:
- Bump version to v3.0.50
- Add html attribute type into Input component
- Adding hover state for Select component


devaneereid (approved):
Looks good!

PR #281, opened 2022-06-14
Feat/v3 list

jaxonjporter:
## Description

This PR adds the list component to amino


src/components/list-item/ListItem.tsx
```diff
@@ -42,39 +81,66 @@ const Icon = styled.img`
...
-      return null;
-    };
     return (
       <AminoListItem
-        className={disabled ? 'disabled' : ''}
-        onClick={() => !disabled && onClick && onClick()}
+        disabled={disabled}
+        selected={selected}
+        withClick={!!onClick}
+        onClick={!disabled ? undefined : onClick}
```
goodpickle, line 135:
```suggestion
        onClick={!disabled && onClick}
```

src/stories/List.stories.tsx
```diff
@@ -15,9 +19,14 @@ export default ListMeta;
 
 export const PrimaryList = () => (
   <List>
```
goodpickle, line 21: What do you think about holding onto a selected state here that could identify the selected list item? This would allow us to test the click functionality.
```
const [selectedListItem, setSelectedListItem] = useState(2);
```

goodpickle (changes requested):

src/stories/List.stories.tsx
jaxonjporter: Thats a great idea

src/components/list-item/ListItem.tsx
jaxonjporter: I changed it back to how it was before.

goodpickle (dismissed):
Looks great! Thanks 👍

PR #285, opened 2022-06-17
fix(button):updated loading icon to display

devaneereid:
## Description
This PR fixes a bug where the loading stage wouldn't render/ the icon couldn't display inside of a button.


devaneereid:
Issue already fixed.

PR #288, opened 2022-06-21
Fix/building v3

Thienhuynh95:
## Description

This PR fixes typescript props onClick type "any" in Button component


src/components/button/Button.tsx
```diff
@@ -300,84 +298,147 @@ export function Button({
...
     }
   }
+
   if (tag === 'div') {
     const buttonProps = {
-      as: tag,
       ...baseProps,
-      ...(props as AsDivProps),
+      ...(props as ButtonProps<'div'>),
+      tag,
```
goodpickle, line 364:
```suggestion
```

src/components/button/Button.tsx
```diff
@@ -300,84 +298,147 @@ export function Button({
...
-        return <TextButton {...buttonProps}>{renderContent()}</TextButton>;
-      case 'link':
-        return <LinkButton {...buttonProps}>{renderContent()}</LinkButton>;
-      case 'secondary':
-      default:
-        return <Secondary {...buttonProps}>{renderContent()}</Secondary>;
-    }
+
+  const buttonProps = {
+    type,
```
goodpickle, line 420:
```suggestion
```

goodpickle (changes requested):
There are just a few additional props that seem to be extra.

src/components/button/Button.tsx
Thienhuynh95: I think the reason for this is to overwrite the default type 'submit' to 'button' if type is not passed in

goodpickle: My bad. Nice catch 😅

goodpickle (approved):

PR #291, opened 2022-06-24
Feat/tooltip

goodpickle:
## Description

This PR adds a tooltip that we can wrap our components with. It is particularly useful to identify why a button is disabled. Most tooltips will not show up on a disabled element. I included a truncate text function that can be used to truncate the subtitle if desired. I didn't include it as part of the tooltip, because automatically truncating the tooltip would seem buggy, especially for a dynamic text. 🤔 

This PR also changes the default text to `body`.

In addition, this PR updates the workflow to publish to `npm` when a PR is merged into `develop` as well as `master`.
We will continue doing `patch` versions for merging into `develop` as a standard along with a `minor` version bump when merging into master (monthly?).

Screen Shot 2022-06-24 at 3 27 35 PM
Screen Shot 2022-06-24 at 3 27 30 PM

## Todo


Thienhuynh95 (approved):

goodpickle:
It looks like we have a design for the tooltip, so I'm going to update this PR to follow that [design](https://www.figma.com/file/WnKnmG7L3Q74hqPsw4rbEE/Amino-2.0?node-id=196%3A10758). Thanks @Thienhuynh95 for finding it 💯

Thienhuynh95 (approved):

PR #292, opened 2022-06-27
Feat/tooltip subtitle react node

goodpickle:
## Description

This PR changes the `Tooltip.subtitle` prop to include a React node. We have a few places in Dashboard where we have a custom `Tooltip.subtitle`.

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "3.0.69",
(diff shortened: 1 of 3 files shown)


jaxonjporter (approved):

goodpickle:
@Thienhuynh95 Thanks for fixing the Select bug with the tooltips showing through 💯

Thienhuynh95 (approved):

PR #314, opened 2022-07-25
Feat/constrain duo tone secondary color

goodpickle:
## Description

This PR constrains the Duotone secondaryColor prop to our theme and adds the table component.

## Todo


src/components/notice/Notice.tsx
```diff
@@ -17,7 +17,7 @@ const AminoNotice = styled.div`
 
   &,
   & > * {
-    font-weight: 500;
+    font-weight: base;
```
Thienhuynh95, line 20:
```suggestion
    font-weight: 500;
```
I think there is no value `base` for the `font-weight` attribute 🤔

goodpickle: Hah! Thanks for catching that. It was a find and replace bug. 🐛

Thienhuynh95 (approved):

Thienhuynh95 (approved):

PR #316, opened 2022-07-27
fix(badge): default font weight & add Dual Currency

goodpickle:
## Description

This PR updates the default font weight to match the designs. This PR also adds the Dual Currency Component and uses it inside the Foreign Currency Expandable Table story.

## Todo

### DIFF
--- a/src/components/currency/Currency.tsx
+++ b/src/components/currency/Currency.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+
+import styled from 'styled-components';
+
+const StyledCurrency = styled.span`
+  font-feature-settings: 'tnum' 1;
+
+  &.negative {
+    color: var(--amino-red-l20);
+  }
+`;
+
+type Props = {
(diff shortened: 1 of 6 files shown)


Thienhuynh95 (approved):

src/components/currency/Currency.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+import styled from 'styled-components';
+
+const StyledCurrency = styled.span`
+  font-feature-settings: 'tnum' 1;
```
Thienhuynh95, line 6: Interesting! This is new to me

Thienhuynh95 (dismissed):

src/components/currency/Currency.tsx
goodpickle: I'm unsure if that is still necessary -> Copied from dashboard 😅

Thienhuynh95 (approved):

PR #320, opened 2022-08-05
Feat/connection map

goodpickle:
## Description

This PR moves a Connection Map from Dashboard to Amino.

## Additional
- Updates the Flag for UK
- Adds the theme.ts to version control (so you don't have to build to use the theme)

## Todo


.gitignore
```diff
@@ -87,4 +87,3 @@ svgReact/svgs
 # Build
 *.map
 build.mjs
-src/styles/constants
```
Thienhuynh95, line 90: I think the reason I put this into `.gitignore` because it's generated by the script and only used for Amino consumer (since Amino can use directly constant from `build-utils` folder). Do we also want to use this constant file inside of Amino ? 🤔

src/components/connection-map/ConnectionMap.tsx
```diff
@@ -0,0 +1,176 @@
...
+    // US to SA (Saudia Arabia) = 1.83
+    return 100;
+  }
+  if (distance < 2.5 && verticalDistance < 70) {
+    // US to AU (Australia) = 2.27
+    return 75;
+  }
+  // US to MG (Madagascar) = 2.52
+  return 50;
+};
```
ckgardner, line 59: This is a pretty cool feature! It looks like this is setting the zoom scale for the map based on how far the connecting line has to stretch. Are the country comments just giving an example of the distance or does this only work for shipments going out of the US?

evad1n:
Is this supposed to look tihs 
image

src/components/connection-map/ConnectionMap.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }
+  if (distance < 1.5 && verticalDistance < 55) {
+    // US to CH (Switzerland) = 1.23
+    return 200;
+  }
+  if (distance < 1.75 && verticalDistance < 60) {
+    // US to KZ (Kazakhstan) = 1.509
+    return 150;
+  }
+  if (distance < 2 && verticalDistance < 65) {
```
Thienhuynh95, line 49: This scale calculation is really cool

src/components/connection-map/ConnectionMap.tsx
goodpickle: The country comments are just examples.

Thienhuynh95 (dismissed):

goodpickle:
> Is this supposed to look this?
> image

@evad1n Thanks! Does this look better?
Screen Shot 2022-08-05 at 3 04 06 PM

.gitignore
goodpickle: @Thienhuynh95 and I chatted about this and determined we could add a comment to the `theme.ts` file which will keep developers from accidentally editing it.

evad1n:
Yeah, that is much better. Is there a way to shift the viewport of the map horizontally so that it has the line in the middle always?

goodpickle:
> Yeah, that is much better. Is there a way to shift the viewport of the map horizontally so that it has the line in the middle always?

That seems like a great improvement, but not one I'm going to pursue at this time. The map scale improvements are already out of scope of the Order Page feature 😅

evad1n (approved):

Thienhuynh95 (dismissed):

Thienhuynh95 (approved):

PR #321, opened 2022-08-05
fix(richradio):updated size & story to match figma

devaneereid:
## Description
This PR adds a size of 64px high - to match new Figma design. RichRadio story has been updated as well. 

## Todo


goodpickle (approved):
Looks great!

PR #324, opened 2022-08-10
Fix/update state for rich radio

jaxonjporter:
## Jira issue
NA

## Description

This PR adds an effect so that on mount the rich radio will display its initial state

## Todo

### DIFF
--- a/src/components/rich-radio/RichRadio.tsx
+++ b/src/components/rich-radio/RichRadio.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, useState } from 'react';
+import React, { ReactNode, useEffect, useState } from 'react';
 import ReactTooltip, { type TooltipProps } from 'react-tooltip';
 
 import { VStack } from 'src/components/stack/VStack';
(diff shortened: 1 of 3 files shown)


goodpickle:
@jaxonjporter This one makes me curious. Where did you experience the problem? 

If possible I would like to reproduce the problem in a Unit test or a Story.

I added a value to the RichRadio.stories.tsx and it appeared to work correctly before your change. 🤷‍♂️

src/components/rich-radio/RichRadio.tsx
```diff
@@ -118,6 +118,10 @@ export const RichRadio = <T extends string>({
 }: RichRadioProps<T>) => {
   const [selectedValue, setSelectedValue] = useState(value);
 
+  useEffect(() => {
+    setSelectedValue(value);
+  }, [value]);
+
```
Thienhuynh95, line 124: It's interesting 🤔 Do you mean to change/adjust value in RichRadio outside of the Rich radio component after the Rich radio is rendered?

jaxonjporter: In the event of a parent fetching data, the initial value will be incorrect. That means on mount the rich radio has the incorrect value. When the value comes in, it needs to reset the state so that it will show the selected option as active.

Thienhuynh95: Gotcha, thanks for clarifying that 😉

goodpickle (approved):

goodpickle:
Thanks Jaxon 💯

Thienhuynh95 (approved):
Looking good!

PR #326, opened 2022-08-12
Test/storyshots

evad1n:
## Description

Adds a sample DOM rendering snapshot test for the Avatar component.

Integrates [storyshots](https://storybook.js.org/addons/@storybook/addon-storyshots) for visual regression testing. 

Storyshots is just generic snapshot testing, we are actually using a subset from [this](https://storybook.js.org/addons/@storybook/addon-storyshots-puppeteer) addon take actual images and compare them.

There were a lot of hacks to get this to work, and even now it sometimes shows false positives. 
> Corporate wants you to find the difference between these pictures
image

I have added some separate scripts to exclude storyshot tests from the standard scripts (and updated vercel to run a test script without it). It requires you to be running the storybook dev server so it can navigate there and take pictures, hence these will have to locally run.

For a11y tests there is also [axe](https://storybook.js.org/addons/@storybook/addon-storyshots-puppeteer) that can be looked into later.

Previously it would take ~90 seconds and fail on ~20 tests. Forcing it to wait for most network requests to be resolved got it to a 99% pass rate, but now it takes upwards of 200 seconds.

---

You can write the tests on a per story basis to get more control, but not every test needs that. I was looking for a way to try and use a the custom specified test, and then fallback to a basic screenshots, but I couldn't find a way within this addon. It can definitely be done by writing a custom crawler with puppeteer, but that is more work.

- [ ] Bump version


package.json
```diff
@@ -129,7 +124,9 @@
     "build:prep:publish": "ts-node build-utils/setupPackage.ts",
     "postbuild:prep:publish": "echo 'done building for publishing'",
     "test": "jest --watchAll",
-    "test:prod": "jest",
+    "test:prod": "jest --testPathIgnorePatterns storyshots",
+    "test:normal": "jest --testPathIgnorePatterns storyshots",
```
Thienhuynh95, line 128:
```suggestion
    "test:prod": "jest --testPathIgnorePatterns storyshots",
```
I think "test:normal" is not necessary

Thienhuynh95 (changes requested):
This is interesting snapshot testing. Good job on implementing this library 💯 
These are huge snapshots to save on git 😆 There are some snapshots that are not good for testing purposes (like the one that needs action click on the button in order to see the modal/coversheet/toaster).  And I feel like it's kinda wasteful to save those snapshot pictures on git. It would be great if we have a way to control, take a snapshot and test it if we need to.

package.json
evad1n: So there is way to do the snapshot on a story-by-story basis. The `puppeteerTest` section of [this](https://storybook.js.org/addons/@storybook/addon-storyshots-puppeteer). Then it would allow us to interact with the  component before we take the screenshot.

.storybook/storyshots.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const beforeScreenshot: ImageSnapshotConfig['beforeScreenshot'] =
+  async page => {
+    // Hide loading spinners as their moving parts interfere with visual regression tests
+    await page.$$eval('.loading', els => {
+      els.forEach(el => {
+        if (el instanceof HTMLElement) {
+          el.style.display = 'none';
+        }
+      });
```
goodpickle, line 25: Good idea 👍

package.json
```diff
@@ -102,19 +110,6 @@
...
-    ],
-    "modulePaths": [
-      "<rootDir>",
-      "build-utils/**/*.ts",
-      "src/**/*.ts"
-    ],
-    "setupFilesAfterEnv": [
-      "./jest/setupJestMock.ts"
-    ]
-  },
```
goodpickle, line 117: Nice cleanup 👍

src/components/avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import * as renderer from 'react-test-renderer';
```
goodpickle, line 2: What do you think about removing `react-test-renderer` and this test? DOM snapshots can be useful in some situations, but I'm hoping to utilize `@testing-library/react` and image snapshots to avoid DOM snapshots.

package.json
```diff
@@ -47,13 +50,18 @@
...
+    "jest-transformer-svg": "^2.0.0",
     "postcss": "8.4.12",
     "postcss-preset-env": "6.7.1",
     "prettier": "2.6.0",
+    "puppeteer": "^16.1.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-hotkeys-hook": "3.4.4",
     "react-is": "^17.0.2",
+    "react-test-renderer": "17.0.2",
```
goodpickle, line 64:
```suggestion
```

goodpickle (changes requested):

src/components/avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { ImageAvatar } from './ImageAvatar';
+import { UserAvatar } from './UserAvatar';
+
+describe('Avatar DOM snapshot tests', () => {
+  it('UserAvatar matches snapshot', () => {
+    const tree = renderer
+      .create(<UserAvatar size="lg" shape="round" />)
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
```
Thienhuynh95, line 17: If I understand correctly,`it` and `test` are the same. But would it be confused if we use both in same project 🤔 I wonder if we should only use just either `it` or `test` function throughout our application or not 🤔 
```suggestion
  test('UserAvatar matches snapshot', () => {
    const tree = renderer
      .create(<UserAvatar size="lg" shape="round" />)
      .toJSON();
    expect(tree).toMatchSnapshot();
  });
```

package.json
Thienhuynh95: Gotcha. Can't wait to see when that can also be implemented 🥳 . This is a good testing implementation for a UI library, really easy to see what has changed when having a picture snapshot of it on the PR. 💯

src/components/avatar/Avatar.test.tsx
evad1n: Whoops, not entirely sure why I used `it`. Must have been habit or something...

src/stories/Spinner.stories.tsx
```diff
@@ -18,3 +18,5 @@ export const BasicSpinner = Template.bind({});
 BasicSpinner.args = {
   size: 32,
 };
+
+BasicSpinner.storyName = 'BasicSpinner';
```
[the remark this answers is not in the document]
evad1n: So the docs are really uninformative for what property the regex actually uses for the storyshots addon. I am using this regex `storyNameRegex: /^((?!.*?BasicSpinner).)*$/,`, but there is also `storyKindRegex`. I have yet to learn the difference, but this managed to work. I will look into if the space works.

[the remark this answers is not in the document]
evad1n: And this is just the exclude that story from visual testing because ... spinners 🔴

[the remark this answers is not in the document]
evad1n: Looks like it works 👍

.storybook/storyshots.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  await new Promise<void>(async resolve => {
+    tryCloseBrowser(resolve);
+  });
+});
+
+describe('Visual regression testing', () => {
+  initStoryshots({
+    suite: 'Puppeteer storyshots',
+    storyNameRegex: /^((?!.*?Basic Spinner).)*$/,
```
Thienhuynh95, line 66: Since this would exclude all spinners components. How do you think about making it to be more generic for all types of Spinner not just `Basic Spinner`?
```suggestion
    storyNameRegex: /^((?!.*?Spinner).)*$/,
```

goodpickle (approved):

.storybook/storyshots.test.ts
evad1n: Good future-facing idea.

Thienhuynh95 (dismissed):
Well done! 💯

Thienhuynh95 (approved):

PR #327, opened 2022-08-15
Test/add test id to component

Thienhuynh95:
## Description

This PR adds the test id attribute to some components to easy to perform component testing in the Dashboard or in other Amino consumer projects

## Todo


package.json
```diff
@@ -97,7 +97,9 @@
     "react-hotkeys-hook": "^3.x",
     "react-is": "^17.x",
     "react-tooltip": "^4.x",
-    "styled-components": "^5.x"
+    "styled-components": "^5.x",
+    "@emotion/react": "^11.9.3",
+    "@emotion/styled": "^11.9.3"
```
goodpickle, line 102: What do you think about making this version more generic? It would be a pain to receive a warning to upgrade except for breaking changes. 🤔 
```suggestion
    "@emotion/react": "^11.x",
    "@emotion/styled": "^11.x"
```

goodpickle (changes requested):

goodpickle (approved):

package.json
evad1n: I'm probably reading [these docs](https://docs.npmjs.com/about-semantic-versioning) wrong, but aren't those the same?

Thienhuynh95: No, `^11.9.3` will ok with any patch `version`, and not `minor` version, so if the library is using `@emotion/react@^11.10.1`, it would not allow and warn a message 'unmet peer dependencies ...'. 
But with `^11.x` will be ok with with `minor` and `patch` version. Does that make sense? 🤔

goodpickle (approved):

package.json
evad1n: 👍 I didn't realize the point was to enable versions of `11.x` lower than `11.9.3`

PR #332, opened 2022-08-18
refactor(file-upload):: make onRemove and uploadedFiles be optional

Thienhuynh95:
## Description

This PR:
- Refactors the file upload component to have `onRemove` and `uploadedFiles` to be optional
- Add `@deprecated` to some variables in `theme.ts`

## Todo

### DIFF
--- a/src/components/file-upload/FileUpload.tsx
+++ b/src/components/file-upload/FileUpload.tsx
@@ -20,8 +20,8 @@ export interface FileUploadProps {
   error?: boolean;
 
   /** Display file info if uploaded file property has data */
-  uploadedFile?: UploadedFileProps | null;
-  onRemove?: (() => void) | null;
+  uploadedFile: UploadedFileProps | null;
+  onRemove?: () => void;
   width?: number;
   helperText?: string;
 }
--- a/src/components/file-upload/FileUpload.tsx
+++ b/src/components/file-upload/FileUpload.tsx
@@ -21,7 +21,7 @@ export interface FileUploadProps {
 
   /** Display file info if uploaded file property has data */
   uploadedFile?: UploadedFileProps;
-  onRemove: () => void;
+  onRemove?: () => void;
   width?: number;
   helperText?: string;
 }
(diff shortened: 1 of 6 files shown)


goodpickle (dismissed):

goodpickle (dismissed):

devaneereid:
Nice refactor Thien!

src/components/file-upload/FileUpload.tsx
```diff
@@ -20,8 +20,8 @@ export interface FileUploadProps {
   error?: boolean;
 
   /** Display file info if uploaded file property has data */
-  uploadedFile?: UploadedFileProps;
-  onRemove: () => void;
+  uploadedFile?: UploadedFileProps | null;
+  onRemove?: (() => void) | null;
```
evad1n, line 24: Why allow null as well as undefined?

Thienhuynh95: The reason for that is if you don't specify it or set it to null, it will not do any action. Does that answer your question? 🤔

evad1n: I guess my question is why would someone set it to null as opposed to just not specifying.

goodpickle, line 24: What do you think about using `null` or `undefined`? `null` is appropriate for props that the Developer would normally expect to use. `undefined` is best for completely optional props that don't affect the functionality of the component.
```suggestion
  uploadedFile?: UploadedFileProps;
  onRemove?: (() => void);
```

goodpickle (changes requested):

src/components/file-upload/FileUpload.tsx
Thienhuynh95: Yah, I got your point. I just have a use case that I need to conditionally trigger the action/object. And I don't love to have a falsy ternary with `undefined`. But I think it's good to have either one or the other. Thanks

Thienhuynh95: I just made up my mind. Thanks for the suggestion

goodpickle (approved):

goodpickle (approved):

PR #334, opened 2022-08-23
fix(toast): give actual unique keys

evad1n:
## Description
Fixes toasts not setting unique key properly. Previously was setting to `Date.now()` inside the map function call, which changed on each render. Now we are using `uuid` to set it to guarantee uniqueness.

## Todo

### DIFF
--- a/src/components/toast/ToastContext.tsx
+++ b/src/components/toast/ToastContext.tsx
@@ -55,12 +55,7 @@ export const ToastContextProvider = ({ children }: Props) => {
               {toasts.map(({ toast, props, uuid }) => {
                 const key = `toast-${toast}-${uuid}`;
                 return (
-                  <Toast
-                    toastKey={key}
-                    key={key}
-                    intent={props?.intent}
-                    data-dog-id={key}
-                  >
+                  <Toast toastKey={key} key={key} intent={props?.intent}>
                     {toast}
                   </Toast>
                 );
--- a/src/components/toast/ToastContext.tsx
+++ b/src/components/toast/ToastContext.tsx
@@ -1,12 +1,7 @@
-import React, {
-  createContext,
-  ReactNode,
-  useCallback,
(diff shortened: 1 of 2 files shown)


src/components/toast/ToastContext.tsx
```diff
@@ -62,12 +52,14 @@ export const ToastContextProvider = ({ children }: Props) => {
...
+              {toasts.map(({ toast, props, uuid }) => {
+                const key = `toast-${toast}-${uuid}`;
                 return (
                   <Toast
-                    toastKey={`toast-${toast}-${Date.now()}`}
-                    key={`toast-${toast}-${Date.now()}`}
+                    toastKey={key}
+                    key={key}
                     intent={props?.intent}
+                    data-dog-id={key}
```
Thienhuynh95, line 62: I think this is unnecessary 🤔 
```suggestion
                    intent={props?.intent}
```

Thienhuynh95 (changes requested):
Looking good to me. Thanks 👍

src/components/toast/ToastContext.tsx
```diff
@@ -62,12 +52,14 @@ export const ToastContextProvider = ({ children }: Props) => {
         <div className="toast-container">
           <div className="toasts-wrapper">
             <AnimatePresence>
-              {toasts.map(({ toast, props }) => {
+              {toasts.map(({ toast, props, uuid }) => {
+                const key = `toast-${toast}-${uuid}`;
```
devaneereid, line 56: Nice fix to make sure each toast is unique!

src/components/toast/ToastContext.tsx
evad1n: whoops

Thienhuynh95 (approved):

goodpickle (approved):

PR #341, opened 2022-09-01
refactor(dropzone): disabled state for dropzone

Thienhuynh95:
## Description

This PR adds `disabled` state for dropzone when no file is selected and a story for it.

## Todo

### DIFF
--- a/src/components/file-upload/FileUpload.tsx
+++ b/src/components/file-upload/FileUpload.tsx
@@ -24,6 +24,8 @@ export interface FileUploadProps {
   onRemove?: () => void;
   width?: number;
   helperText?: string;
+  /** @desc This `disabled` state only apply to dropzone when no file is selected */
+  dropzoneDisabled?: boolean;
 }
 
 type DropzoneWrapper = {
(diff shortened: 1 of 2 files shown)


devaneereid:
Looks good Thien!

src/components/file-upload/FileUpload.tsx
```diff
@@ -115,6 +131,7 @@ export const FileUpload = ({
   onRemove,
   uploadedFile,
   width,
+  dropzoneDisabled,
```
goodpickle, line 134: What do you think about using `disabled` to stay consistent with other amino components?
```suggestion
  disabled,
```

goodpickle (changes requested):

src/components/file-upload/FileUpload.tsx
Thienhuynh95: That was my original intention. I thought about that, but then I think it will be confused with `disabled` props in dropzoneOptions. I want `disabled` prop only apply on the dropzone when there is no file selected. That's why I try the name to be more specific 🤔 What do you think?

goodpickle: I'm confused. Let's chat about this.

goodpickle: We chatted and determined that `dropzoneDisabled` is a good name, however we are removing the `dropzoneOptions.disabled` so we don't have two props disabling dropzone.

goodpickle (dismissed):

goodpickle (approved):

PR #345, opened 2022-09-06
Fix/badge props

evad1n:
## Description

Badge prop for `rounded` was reversed. As a result, consumers will most likely needed to set `rounded` prop to have the same effect.

Some QOL storybook changes

## Todo


goodpickle (approved):

devaneereid:
Looks great Will!

goodpickle (approved):

PR #347, opened 2022-09-09
feat(hsection): add HSection & collapsable ability

Thienhuynh95:
## Description

This PR adds `HSection` with collapsable ability and rename `Section` component to `VSection`

## Todo


evad1n (approved):
I think it might be nice to see an HSection/Vsection story with multiple parts as children. That would reflect most of our use cases, especially for the collapsible one. What do you think?

Thienhuynh95:
> I think it might be nice to see an HSection/Vsection story with multiple parts as children. That would reflect most of our use cases, especially for the collapsible one. What do you think?

Thanks for the feedback. Which multiple parts you would like to see? Do you mean all states (collapse, expand, long subtitle) in one story?

evad1n (approved):

evad1n:
Just a p

> > I think it might be nice to see an HSection/Vsection story with multiple parts as children. That would reflect most of our use cases, especially for the collapsible one. What do you think?
> 
> Thanks for the feedback. Which multiple parts you would like to see? Do you mean all states (collapse, expand, long subtitle) in one story?

Just something to show how it looks in common cases, so surrounded by other sections as well where you can collapse each section.

Thienhuynh95:
Sounds good, I'll put that in the next PR. Thanks 😄

PR #349, opened 2022-09-09
fix: native datetime placeholder is hidden correctly now

evad1n:
## Description

Given a `min` and `max` for a date input, it will render a native placeholder, which interfered with our label. This fixes that (thanks @Thienhuynh95 for the fix)

Old:
image

The new one just hides that placeholder until we focus.
image

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "3.2.14",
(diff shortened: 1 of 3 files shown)


Thienhuynh95 (approved):
Looks great!!! 😉

Thienhuynh95 (approved):

PR #350, opened 2022-09-12
Feat/button loading spinner colors

evad1n:
## Description

- Adds some extra spinner colors roughly based on intent
- Fixes button loading spinner color being blue for certain variants
- Improves spinner and button stories

When editing the `Button` component, I saw a lot of logic for switching the tag between `a`, `div`, and `button`. I went back and looked at the reason and it seemed like it was for the `onClick` event type. I consolidated it to try and reduce the redundancy, and it seems like it infers the type correctly still.

Also not sure on how necessary/useful the other spinner colors will be. I did them before the button, and then realized having the same spinner color as the button didn't look that good compared to just white.

## Todo


src/types/Color.ts
```diff
@@ -24,3 +24,11 @@ export const colorContrasts = [
...
 type ColorPrefix = typeof colorPrefixes[number];
 export type Color = `${ColorPrefix}-${ColorContrast}` | 'black';
+
+export type ThemeColor =
+  | 'primary'
+  | 'secondary'
+  | 'info'
+  | 'success'
+  | 'warning'
+  | 'danger';
```
Thienhuynh95, line 34: I think we already have the type for this. What do you think about pick the type from this `src/types/Intent.ts`?

src/components/spinner/Spinner.tsx
```diff
@@ -68,13 +71,26 @@ const AminoSpinner = styled.span<SpinnerProps>`
...
+  &.danger {
+    border-top-color: ${theme.redBase};
+  }
+  &.warning {
+    border-top-color: ${theme.orangeBase};
+  }
 `;
 
-type Color = 'black' | 'white';
+export type SpinnerColor = 'black' | 'white' | ThemeColor;
```
Thienhuynh95, line 89: What do you think about picking type from the list of color in `src/types/Intent.ts` instead of creating a new type?

src/types/Color.ts
evad1n: If you mean use `Pick` or `Omit`, then yeh I can do that. But I want a type that isn't as confusing as `Intent`. Some components implement `error`, and the others do `danger` and it is very confusing.

Thienhuynh95: Yes, using `Pick` and `Omit` is what I meant. I agree it's confusing, that's why we `Pick` or `Omit` from the `Intent`. I think it would be great if we can make it consistent and don't have 2 types that are the same, but just one has more than the other. If we would like to modify the type to either `danger` or `error`, we should update the `Intent` but not create a new type. Does that make sense?

evad1n: I didn't want to modify `Intent` because that would cause a lot of breaking changes for any consumers. I think we should do it at some point though. Is that what you are suggesting?

`export type ThemeColor = Exclude<Intent, 'default' | 'error'>;`

Thienhuynh95: What I meant is doing that for the `SpinnerColor` type and removing the `ThemeColor` type. Then we can remove the `error` out of the `Intent` as we go. What do you think?
```
export type SpinnerColor = 'black' | 'white' | Exclude<Intent, 'default' | 'error'>;
```

evad1n: That works for this case. I was anticipating a type to be used in the future that would simplify the `Intent` prop so we don't use types we are eliminating in the future. I am fine with that if we update the `Intent` prop sooner rather than later. 👍

Thienhuynh95 (approved):
Looks awesome! Well done!

Thienhuynh95 (approved):

PR #352, opened 2022-09-13
FIX: Allow div props on vstack

ckgardner:
## Description
There was an issue when I updated the docs to use the latest version of Amino. The Vstack typescript no longer allowed a tabIndex prop. This PR adds typing to allow that to happen.

### DIFF
--- a/src/components/stack/Stack.tsx
+++ b/src/components/stack/Stack.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 
 import styled from 'styled-components';
-import { Any } from 'ts-toolbelt';
 
 import { GridAlignment, GridSpacing } from './GridSpacing';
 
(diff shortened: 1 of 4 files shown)


goodpickle (approved):

src/components/stack/Stack.tsx
```diff
@@ -20,21 +28,10 @@ export type StackProps = {
...
-  alignment,
-  children,
-}: StackProps) => {
-  return (
-    <StyledStack className={className} spacing={spacing} alignment={alignment}>
-      {children}
-    </StyledStack>
-  );
+export const Stack = ({ children, ...otherProps }: StackProps) => {
+  return <StyledStack {...otherProps}>{children}</StyledStack>;
```
devaneereid, line 36: Looks great with how the props are passed 🎉 Nice job Cade!

PR #355, opened 2022-09-16
fix: TextAvatar inline styling

evad1n:
## Description

When trying to fix the store switcher in dashboard I ran into a serious scrolling performance issue. To make a (really) long story short, the issue was actually the styled component `TextAvatar` it was rendering. styled-components gave me a warning in the console for generating over 200 classes and a slightly cryptic fix.

TLDR: CSS class generation is computationally expensive, especially in large lists. Inline styles are quick. 

Also made the TextAvatar story more ... colorful 😄 

Console warning in question:
```
instrument.js?b344:107 Over 200 classes were generated for component styled.div with the id of "sc-jGprRt".
Consider using the attrs method, together with a style object for frequently changed styles.
Example:
  const Component = styled.div.attrs(props => ({
    style: {
      background: props.background,
    },
  }))`width: 100%;`

  <Component />
```

## Todo

### DIFF
--- a/src/components/text-avatar/TextAvatar.tsx
+++ b/src/components/text-avatar/TextAvatar.tsx
@@ -1,14 +1,25 @@
 import React from 'react';
 
 import { theme } from 'src/styles/constants/theme';
-import styled from 'styled-components';
+import styled, { CSSProperties } from 'styled-components';
 
-type GraidentSquareType = {
+type GradientSquareType = {
   gradientStart: string;
   gradientEnd: string;
 };
 
-const GradientSquare = styled.div<GraidentSquareType>`
+// Used in store list, and when too many classes are generated styled-components complains. This is their recommended solution to make the changes styles inline, rather than generating a CSS class for each one.
+const GradientSquare = styled.div.attrs<GradientSquareType>(
+  (props): { style: CSSProperties } => ({
+    style: {
+      background: `linear-gradient(
+      137deg,
+      ${props.gradientStart},
+      ${props.gradientEnd}
+    )`,
+    },
+  })
+)<GradientSquareType>`
   width: 32px;
   height: 32px;
   border-radius: ${theme.radius};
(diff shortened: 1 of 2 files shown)


src/components/text-avatar/TextAvatar.tsx
```diff
@@ -1,14 +1,25 @@
...
+  (props): { style: CSSProperties } => ({
+    style: {
+      background: `linear-gradient(
+      137deg,
+      ${props.gradientStart},
+      ${props.gradientEnd}
+    )`,
+    },
+  })
+)<GradientSquareType>`
```
Thienhuynh95, line 22: Do you have any purpose behind just putting a portion of the style and not the part in the back tick here?

src/components/text-avatar/TextAvatar.tsx
```diff
@@ -1,14 +1,25 @@
 import React from 'react';
 
 import { theme } from 'src/styles/constants/theme';
-import styled from 'styled-components';
+import styled, { CSSProperties } from 'styled-components';
 
-type GraidentSquareType = {
+type GradientSquareType = {
```
goodpickle, line 6: 🤓 I love seeing the typo's fixed

goodpickle (approved):
Really interesting fix. This would be a good one to teach about in our Planning meeting.

src/components/text-avatar/TextAvatar.tsx
evad1n: That's how styled-components said to do it 🤷 . Only inline the parts that change and keep the rest outside to maintain the normal CSS syntax. Inside the `attrs` it is JS CSS syntax (camelCase).

jaxonjporter (approved):
That is interesting, but nice find.

src/components/text-avatar/TextAvatar.tsx
Thienhuynh95: Interesting 😆

Thienhuynh95 (approved):
Good fix!!! Looks awesome

PR #357, opened 2022-09-19
fix: revert d3-geo to pre type:module version

goodpickle:
## Description

This PR fixes a Nextjs build error by reverting the `d3-geo` package down to 2.0.2 which is before `d3-geo` started compiling down to using the `import` syntax instead of the `require` syntax.

### Nextjs Build error:
```
require() of /vercel/path0/node_modules/d3-geo/src/index.js from /vercel/path0/node_modules/@amino-ui/core/components/connection-map/ConnectionMap.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
```

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "3.2.20",
+  "version": "3.2.21",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:amino-ui/core.git",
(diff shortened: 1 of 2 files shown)


Thienhuynh95 (dismissed):
Look awesome 😉

Thienhuynh95 (approved):

PR #363, opened 2022-09-26
feat(table-row): allow expand with children

goodpickle:
## Description

This PR allows expanding the Expandable Table Row when there are children. In addition, the click event is now on the table row instead of the button.

## Todo


Thienhuynh95 (dismissed):
Awesome!

ckgardner (commented):
Looks good!

evad1n (approved):

PR #371, opened 2022-10-13
Style/fix button w long content inside short wrapper

Thienhuynh95:
## Description

This PR adds `white-space: nowrap;` to the button to always display text in one line since the button height would never change but the width does.
image

## Todo


jaxonjporter (approved):
Good fix.

PR #372, opened 2022-10-13
build(actions): add back registry-url

evad1n:
## Description

As stated [here](https://github.com/actions/setup-node/blob/main/action.yml)

The last tag failed to publish, so this should trigger that


Thienhuynh95 (approved):
Looks good!

goodpickle (approved):

PR #373, opened 2022-10-20
fix: overflow auto when fully expanded

evad1n:
## Description

Fixes an issue with the `Collapse` component where `overflow: hidden` was causing box shadows (and anything outside the components bounding box) to be hidden.

Also made the Collapse story slightly nicer.

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "3.2.29",
+  "version": "3.2.30",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:amino-ui/core.git",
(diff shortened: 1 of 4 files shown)


goodpickle (approved):

jaxonjporter (approved):
I like your solution.

evad1n:
Ha, it was the least code solution

jaxonjporter:
> Ha, it was the least code solution

I dont know, padding: 2px probably would have been the least code lol

evad1n:
> > Ha, it was the least code solution
> 
> I dont know, padding: 2px probably would have been the least code lol

I wouldn't consider that a good solution 😆

PR #378, opened 2022-11-09
Feat/dialog subtitle

evad1n:
## Description

Adds optional subtitle prop to dialog.

Also changes close icon to 24px to closer match design.

## Todo


goodpickle (approved):
Looks great!

goodpickle (approved):

PR #386, opened 2022-12-01
Feat/design tweaks

evad1n:
## Description

This PR tweaks some designs to enable easier completion of a task I am working on.

- Makes Dialog label able to be a react node
- Matches badge to the current figma designs (besides the colors because the color situation is not so good)
- Text type button's hover effect works correclty
- Dialog actions don't use HStack so they can be variable widths (like in all the designs)
- Banner icons will align to top if there is only content (this was the case in a lot of places already and it looked not so good when the icon was centered vertically).

## Todo


goodpickle (approved):

goodpickle (dismissed):

Thienhuynh95 (approved):
Looks good.

Thienhuynh95 (approved):

PR #390, opened 2022-12-05
Feat/new color generation process

Thienhuynh95:
## Issue
The design team has recently updated the color set palette to the new one (no longer use the intensity scale from L80 to D80, now it's from 100 - 1000), and it will be time-consuming to migrate to the new scale.  

## Overview 
This PR would ease the process by:
- Just exporting the color component in Figma and placing them inside the folder `build-utils/css/colorSvgs`. 
  image
- Running the command `yarn build:colors` will generate the color constant that can be used in `theme.ts` based on the svg loaded from the folder mentioned above. 
  image
- If you generated color constants and you forget to use/import any constant in the `theme.ts`, and when you run `yarn build` or `yarn build:theme`, it would also check and terminate the process intentionally to tell you which file is not being used.
  image

## Description

This PR:
- Update README for flow of generating color from Figma.
- Adds a new script to build color constants out of the svgs downloaded from Figma.
- Apply lint fix on some places in the `build-utils` that Eslint couldn't reach.
- Adds tests for the whole process.
- Adds and updates the new color key and also adds a tag `@deprecated` to the old color scale.

## Todo


build-utils/css/utils/logging.ts
```diff
@@ -0,0 +1,4 @@
+export const logging = (message: string) => {
+  // eslint-disable-next-line no-console
+  console.log(message);
+};
```
evad1n, line 4: It might be easier to do something like this: https://stackoverflow.com/a/50101673/15412237

evad1n (commented):
Is it necessary to include the rest of the SVG content (like the `path` info)? It seems the majority of the svg color file content is the path info and it could save a lot of space. Also is it even necessary to keep the SVGs at all if we only need them once to generate?

build-utils/css/README.md
```diff
@@ -1,19 +1,129 @@
...
+    2. Put all of the exported svgs into `build-utils/css/colorSvgs` folder, and make sure the component name is in CamelCase (Blue100.svg, Blue200.svg, ...).
+    3. **SKIP THIS STEP if the variable name doesn't change. If the variable from the Figma and the current variable are different (like the variable name before was `BlueL80` and now it become `Blue100`).**
+        - Look into `build-utils/css/utils/generateColorConstantFromSvgs/generateFileContent.ts` file.
+        - Adjust the mapping configuration for the color in the mapping array `deprecatedIntensityMapping`. In the object, the key on the left would be the legacy color suffix and the value on the right would be the new one. This mapping configuration would also add the legacy colors down below and add the `@deprecated` tag above it after the content of the new color.
+        - You can also adjust the content of constant `content` in the function `generateFileContent` to modify what's being generated if needed (Like add jsdocs or change text or adding more deperecated color).
```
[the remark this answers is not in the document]
Thienhuynh95: Good catch! Thanks.

evad1n:
I see your point, however if it is only a color, then maybe it isn't necessary to view it as an svg right? Just the hex code history should be enough.

Thienhuynh95:
Hmn, kinda. I think the main point of my thought is to see the svgs actually have a color, not broken, and also make sure you are not selecting and grouping more than 1 color to export. Because if you accidentally do so, it would mess up the color code. Because this whole process is just downloading a whole set of icons on Figma and pasting them into a folder, then running the script to generate code. I assume that no one would click every single svg and preview what they downloaded 😅

evad1n (dismissed):
💯

build-utils/css/utils/logging.ts
Thienhuynh95: I'll have this on my next PR. Thanks

PR #391, opened 2022-12-07
Stories/file based routing

evad1n:
## Description

This PR restructures the stories and related components to be next to the actual component in the file structure. 

Also removes the default React import following https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

## Todo


src/components/alert-dialog/__stories__/Alert.stories.tsx
```diff
@@ -1,27 +1,22 @@
-import React from 'react';
-
 import { Meta, Story } from '@storybook/react/types-6-0';
 import { AlertContextProvider } from 'src/components/alert-dialog/AlertContext';
-import { withDesign } from 'storybook-addon-designs';
 
 import { AlertConsumer } from './AlertConsumer';
```
Thienhuynh95, line 5: Since you already have a good solution for grouping stories in same folder, what do you think about making this to be more consistent by moving them all out of the `__stories__` folder. And if inside of that `__stories__` folder has a component that is just being used internally for the stories, then adding a `_` prefix? 🤔

.storybook/main.ts
```diff
@@ -1,7 +1,35 @@
...
+
+  // Will order stories alphabetically by file name. If a title is specified it will take priority over the file name.
+  const alphabetical = stories.sort((a: string, b: string) => {
+    const storyNameA = getTitle(resolve(__dirname, a)) || a;
+    const storyNameB = getTitle(resolve(__dirname, b)) || b;
+
+    return basename(storyNameA).localeCompare(basename(storyNameB));
+  });
+
+  return alphabetical;
```
Thienhuynh95, line 20: I love it.

Thienhuynh95 (approved):
Looks good 💯

Thienhuynh95 (approved):

PR #393, opened 2022-12-09
Fix/new color redesign

evad1n:
## Description

This PR changes all color variables to use the camelCase format and access the `theme` object.

- Fixes Orange color (was cyan by accident)
- Updates scripts for new color formats
- Small story improvements when looking through each story (e.g. some had controls that were stateless with null `onChange` so they were not interactable)

## Todo

---
For posterity, the VSCode color replace regex is 
`(['"]\w+)-(\d{3}['"])`
and
`$1$2`

This does not account for 4 value color shades (`orange1000` or `gray1200`), but those are hardly used :)

Also if you need to use this make sure to look for any `var(--amino-` interaction with colors as it will silently break.


Thienhuynh95 (dismissed):
Looks good. Thanks for the updating color to camelCase.

Thienhuynh95 (approved):

PR #396, opened 2022-12-13
Fix/dialog height flex

evad1n:
## Description

This PR fixes the dialog so the content will flex grow and push the actions to the end.

Also:
- fixes border color to match old one
- puts max height on dialog parent instead of content
- dialog story has height control

## Todo


goodpickle (approved):

Thienhuynh95 (approved):
Looks good!

PR #398, opened 2022-12-22
Feat/color mapping change

Thienhuynh95:
## Description

After making the changes to the color, and mapping the old colors with the new ones, the designers realized that the mapping they gave previously is not quite right. So they would like to update the mapping a little bit.

This PR would update the color mapping:

Old mapping.                      |  New mapping
:-------------------------:|:-------------------------:
CleanShot 2022-11-10 at 11 03 03@2x | CleanShot 2022-12-19 at 10 01 58@2x

## Todo


evad1n (dismissed):
🌈

goodpickle (approved):

evad1n (approved):

PR #402, opened 2023-01-03
Fix/connection map loader

evad1n:
## Description

This PR adds a skeleton loader to the connection map to prevent a jarring transition.

- fixes the publish action to not use `develop` anymore
- sorts the build generated `all.ts` file
- add tsconfig rule for index access checks

## Todo


goodpickle (approved):

Thienhuynh95 (dismissed):
Looks good! 💯

src/components/slider/Slider.tsx
```diff
@@ -100,7 +100,7 @@ export const Slider = ({ min, max, value, step, onChange }: SliderProps) => (
     <StyledSlider
       max={max}
       min={min}
-      onValueChange={([val]) => onChange(val)}
+      onValueChange={([val]) => val !== undefined && onChange(val)}
```
Thienhuynh95, line 103:
```suggestion
      onValueChange={([val]) => !!val && onChange(val)}
```
I think you can do this instead 🤔

tsconfig.json
```diff
@@ -26,7 +26,8 @@
     ],
     "esModuleInterop": true,
     "downlevelIteration": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "noUncheckedIndexedAccess": true
```
Thienhuynh95, line 30: I love this tsconfig rule 💯

Thienhuynh95 (changes requested):

src/components/slider/Slider.tsx
evad1n: It is a number, which could be 0 so I need to explicitly check for only undefined.

tsconfig.json
evad1n: I know 😄

src/components/slider/Slider.tsx
Thienhuynh95: I just looked deeper into it, the type of `[val]` claimed that it's always a number, would it ever be `undefined`?

evad1n: It is an indexed access (array destructure) so it can always be undefined

Thienhuynh95: Got it. All of these changes are because you turned on the ts rule for index accessing and it's stricter checking 😆 That's nice!!!

Thienhuynh95 (approved):
Everything looks great! Thanks

Thienhuynh95 (approved):

PR #403, opened 2023-01-04
Feat/textarea

Thienhuynh95:
## Description

Storybook preview domain: https://amino-njmc87cx4-zonos.vercel.app/?path=/story/amino-textarea--default

Text area in the Figma Catalog design
image

This PR adds a component for Textarea:
- Regular:
  image
- With help text:
  image
- Error with help text:
  image
- Regular error state:
  image

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "3.3.7",
+  "version": "3.3.8",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:amino-ui/core.git",
   "author": "Joshua Beitler <joshbeitler@gmail.com>",


evad1n:
I think this readonly one might look better if it wasn't highlighted (blue outline) and maybe was disabled

image

Like this disabled one:
https://mui.com/material-ui/react-text-field/#form-props

Thienhuynh95:
> I think this readonly one might look better if it wasn't highlighted (blue outline) and maybe looked more disabled (opacity?)
> 
> image
> 
> Like this disabled one: https://mui.com/material-ui/react-text-field/#form-props

I thought the concept of `readonly` is it's just the regular component but just not editable 🤔 🤷‍♂️ 
According to the Material UI demo you gave, I hardly tell the UI difference between the `default`, `required` and the `readonly` one. The textarea also has the `disabled` state though, maybe that's the one you meant?
image

evad1n:
Yeh I guess there is read only and disabled, oops 👍

evad1n (approved):

evad1n (approved):

PR #405, opened 2023-01-10
update/All icons

ckgardner:
## Description

This PR updates all of the icons that have changed in Figma. Solid icons have been removed and white icons have been added. The reason white icons were needed is because they have a larger thickness than the black icons in order to stand out on a black background. 

## Todo


evad1n:
So from the following image it looks like the white icon is the one with less thickness, which means it is supposed to be used on a white background, while the normal one is slightly thicker, which should be used for dark backgrounds?
image

evad1n (dismissed):
Went through and looked at some of the components and it looks good 👍

Thienhuynh95:
@evad1n According to the Figma, `White` icons are being located in the black background, so I assume `White` will be used on DarkMode, and the other rest is regular mode 👀

evad1n:
Huh, that makes sense I think. Some confusing naming though.

Thienhuynh95:
😆 You don't say 🙄

Thienhuynh95 (dismissed):
Looks great!

goodpickle (approved):

PR #412, opened 2023-01-17
feat/more icons

ckgardner:
## Description

This PR brings in the latest product icons from Figma.

## Todo


src/all.ts
```diff
@@ -173,8 +173,11 @@ import './icons/CheckCircleWhiteIcon';
 import './icons/CheckmarkDuotoneIcon';
 import './icons/CheckmarkIcon';
 import './icons/CheckmarkWhiteIcon';
+import './icons/CheckoutBlack1Icon';
+import './icons/CheckoutDuotone1Icon';
```
goodpickle, line 177: What is going on here? Is the `1` intentional?

goodpickle (changes requested):

src/all.ts
ckgardner: Oh I need to rerun the command. For some reason, Amino was exported duplicates of the product icons so I had to remove 9 icons, but I did that afterwards. I'll fix it.

goodpickle (dismissed):

goodpickle (approved):

PR #413, opened 2023-01-18
fix/product icons

ckgardner:
## Description

This PR adds the new product icons. They are named the same as the old product icons and will replace them directly. The old product icons have been renamed to their shape name (sqaure, circle, etc) if we still want to use them.  

## Todo

- [] Bump version and add tag


goodpickle (changes requested):
We need to chat about versioning Icons. I wouldn't think we would want multiple versions. Do we have other versions in Amino?

ckgardner:
> We need to chat about versioning Icons. I wouldn't think we would want multiple versions. Do we have other versions in Amino?

Not sure what you mean exactly. Don't we need to version in order to tell npm there is an update on our projects?🤔 Or are you just referring to the back to back versions?

goodpickle:
> > We need to chat about versioning Icons. I wouldn't think we would want multiple versions. Do we have other versions in Amino?
> 
> Not sure what you mean exactly. Don't we need to version in order to tell npm there is an update on our projects?🤔 Or are you just referring to the back to back versions that shouldn't happen?

Oh, I am talking about the Icon Version. It is weird to have `./icons/CheckoutV2Icon`. What is the purpose? I would expect we would want only 1 version of each icon 🤔

ckgardner:
> > > We need to chat about versioning Icons. I wouldn't think we would want multiple versions. Do we have other versions in Amino?
> > 
> > 
> > Not sure what you mean exactly. Don't we need to version in order to tell npm there is an update on our projects?🤔 Or are you just referring to the back to back versions that shouldn't happen?
> 
> Oh, I am talking about the Icon Version. It is weird to have `./icons/CheckoutV2Icon`. What is the purpose? I would expect we would want only 1 version of each icon 🤔

Oohhh gotcha. Connor said there are 2 sets of product icons right now. The new ones will eventually replace the old ones and will would remove the V2. Is there another way you would like me to recommend to name them?

Thienhuynh95:
@ckgardner We have legacy folder to store the old icon, what do you think about putting the old icons into `legacy` folder and the new one stay in the main folder?

goodpickle:
> > > > We need to chat about versioning Icons. I wouldn't think we would want multiple versions. Do we have other versions in Amino?
> > > 
> > > 
> > > Not sure what you mean exactly. Don't we need to version in order to tell npm there is an update on our projects?🤔 Or are you just referring to the back to back versions that shouldn't happen?
> > 
> > 
> > Oh, I am talking about the Icon Version. It is weird to have `./icons/CheckoutV2Icon`. What is the purpose? I would expect we would want only 1 version of each icon 🤔
> 
> Oohhh gotcha. Connor said there are 2 sets of product icons right now. The new ones will eventually replace the old ones and will would remove the V2, but both icons are used right now. Is there another way you would like me to recommend to name them?

I'm a little confused. Are you saying we don't want to replace all of the Product Icons with the new version? 

I would definitely prefer to have only one version of Icon out there. However, if for some reason we can't, I would rename the old ones with a prefix of `Legacy`, and then update the projects using the old ones.

ckgardner:
> > > > > We need to chat about versioning Icons. I wouldn't think we would want multiple versions. Do we have other versions in Amino?
> > > > 
> > > > 
> > > > Not sure what you mean exactly. Don't we need to version in order to tell npm there is an update on our projects?🤔 Or are you just referring to the back to back versions that shouldn't happen?
> > > 
> > > 
> > > Oh, I am talking about the Icon Version. It is weird to have `./icons/CheckoutV2Icon`. What is the purpose? I would expect we would want only 1 version of each icon 🤔
> > 
> > 
> > Oohhh gotcha. Connor said there are 2 sets of product icons right now. The new ones will eventually replace the old ones and will would remove the V2, but both icons are used right now. Is there another way you would like me to recommend to name them?
> 
> I'm a little confused. Are you saying we don't want to replace all of the Product Icons with the new version?
> 
> I would definitely prefer to have only one version of Icon out there. However, if for some reason we can't, I would rename the old ones with a prefix of `Legacy`, and then update the projects using the old ones.

ckgardner:
> > > > > We need to chat about versioning Icons. I wouldn't think we would want multiple versions. Do we have other versions in Amino?
> > > > 
> > > > 
> > > > Not sure what you mean exactly. Don't we need to version in order to tell npm there is an update on our projects?🤔 Or are you just referring to the back to back versions that shouldn't happen?
> > > 
> > > 
> > > Oh, I am talking about the Icon Version. It is weird to have `./icons/CheckoutV2Icon`. What is the purpose? I would expect we would want only 1 version of each icon 🤔
> > 
> > 
> > Oohhh gotcha. Connor said there are 2 sets of product icons right now. The new ones will eventually replace the old ones and will would remove the V2, but both icons are used right now. Is there another way you would like me to recommend to name them?
> 
> I'm a little confused. Are you saying we don't want to replace all of the Product Icons with the new version?
> 
> I would definitely prefer to have only one version of Icon out there. However, if for some reason we can't, I would rename the old ones with a prefix of `Legacy`, and then update the projects using the old ones.

Yeah, he just said they aren't quite ready to switch all of them, but wanted to start with the docs. So moving to the legacy folder as Thien suggested would work.

goodpickle (dismissed):

Thienhuynh95 (approved):
💯

goodpickle (approved):

PR #423, opened 2023-01-31
refactor: strongly type for countryselect

Thienhuynh95:
## Description

This PR:
- Improve `CountrySelect` and `CountryMultiSelect` to provide more strongly typed data.
**Ex**: If some strongly typed `RandomCountryCode` variables are passed in `countryOptions`, `onChange` and `value` should give the correct type based on the `countryOptions` variables, which is `RandomCountryCode` and not just primitive `string` type.
image
image

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "3.4.14",
+  "version": "3.4.15",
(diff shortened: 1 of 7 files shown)


goodpickle (approved):
That's what I'm talking about! 💯

evad1n (approved):
.'//

PR #424, opened 2023-01-31
Feat/ripple click

evad1n:
## Description

This PR adds a new Ripple effect to buttons. It was based off the material UI ripple and can be added to any component using the useRipple hook. See the story for an example.

Also fixes the collapse to use a layout effect so the scrollbar doesn't go crazy.

## Todo


Thienhuynh95 (approved):
🔢 Flawless

.eslintrc.js
```diff
@@ -58,6 +58,12 @@ module.exports = {
     '@typescript-eslint/no-explicit-any': 'error',
     camelcase: 'off',
     'no-unused-vars': 'off', // Disable the base rule it can report incorrect errors
+    'no-unused-expressions': [
+      'error',
+      {
+        allowShortCircuit: true,
```
goodpickle, line 64: 🤔 Where did this rule break?

goodpickle (changes requested):

src/components/button/Button.tsx
```diff
@@ -317,46 +319,46 @@ type ButtonBase = {
   size?: Size;
   tabIndex?: number;
   type?: 'button' | 'reset' | 'submit';
-  /**
-   * Color for the spinner when in a loading state
-   */
+  /** Color for the spinner when in a loading state */
   spinnerColor?: SpinnerProps['color'];
+  /** Disable ripple effect */
+  noRipple?: boolean;
```
goodpickle, line 325: What do you think about alphabetizing this prop?

goodpickle (changes requested):

.eslintrc.js
evad1n: The original code I wrote was doing something with possible nonexistent event handlers. Not sure what the reason was I couldn't do the optional call like I am now. Looks like it doesn't break anymore so I will remove it.

src/components/button/useRipple.ts
```diff
@@ -0,0 +1,76 @@
...
+          rippleHandlers[eventName](ev);
+        },
+      }));
+
+      return wrappedEvents.reduce((acc, current) => {
+        acc[current.name] = current.handler;
+        return acc;
+      }, {} as IRippleEventHandlers);
+    };
+  }, [handleMouseDown]);
```
evad1n, line 70: Originally this was combining many event handlers, but I ended up only having the one. It looks more complicated than it needs to now, but if similar logic needs to be done in the future this will be valuable to keep around (in my opinion).

goodpickle (dismissed):

goodpickle (approved):

PR #429, opened 2023-02-08
fix: collapse initial state animation and add to story to show

evad1n:
## Jira issue

[MdxCollapse animates open instead of just being open right on page load](https://myzonos.atlassian.net/browse/DOCS-292)

## Description

This PR fixes an issue where the Collapse would animate the initial state if it started open. I also made it `opacity:0` because if it ever had borders and such it would still show while collapsed.

## Todo

### DIFF
--- a/src/components/collapse/Collapse.tsx
+++ b/src/components/collapse/Collapse.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useLayoutEffect, useState } from 'react';
+import { ReactNode, useEffect, useLayoutEffect, useRef, useState } from 'react';
 
 import { useResizeAware } from 'src/utils/useResizeAware';
 import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


src/components/collapse/Collapse.tsx
```diff
@@ -43,23 +48,40 @@ export const Collapse = ({
     }
   }, [isExpand, setHeight, sizes.height, hideOverflow]);
 
+  useEffect(() => {
+    const { current } = el;
```
Thienhuynh95, line 52: The initial value for this `el` ref is `null`. Does it still allow you to spread `el` even that it's possibly null?

evad1n: Yes, because the `current` is the typed value, not the el. The passed generic type only affects what it holds.

Thienhuynh95 (dismissed):
💯

ckgardner (approved):
Thank you for solving my problems

PR #430, opened 2023-02-08
test: playwright playground

Thienhuynh95:
## Description

This PR adds PlayWright to the project for the experiment.
In this PR:
- adds github action for running the visual test. After the storybook has been deployed and the Vercel preview url is ready, the test will be run on that preview amino url. If it ran successfully, it will use the cache and not redownload the browser dependencies.
image

- adds a visual test for the `Confirmation` stories.

## Todo


goodpickle (approved):

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,50 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test_setup:
+    name: Test setup
+    runs-on: macos-latest
```
evad1n, line 10: Why use macos-latest here and macos-12 later?

Thienhuynh95: Thanks for catching that. I realized the second job's OS is important, changed it, and forgot to update the OS of the first job 😆 . But the `test_setup`'s OS would not affect the test or anything since it is only the placeholder for the next actual test job. I'll update the yml on my next PR.

evad1n (approved):

PR #431, opened 2023-02-09
Refactor/collapse

evad1n:
## Description

This PR refactors the Collapse component completely. I based this approach on [material-ui's approach](https://github.com/mui/material-ui/blob/next/packages/mui-material/src/Collapse/Collapse.js) (you may notice a pattern here, I mean their stuff is pretty good so I'm not trying to reinvent the wheel here). I developed this inside the documentation repo so it should definitely work there, and have full feature parity with the old Collapse.

- Changes a lot of collapse prop names to be consistent.
- Adds `plain` button intent type, which is generally used for icon buttons.

## Todo


goodpickle (approved):

Thienhuynh95 (approved):
Good refactor! 💯

Thienhuynh95 (approved):

src/components/button/Button.tsx
```diff
@@ -455,6 +460,13 @@ export function Button<T extends GroupTag = 'button'>({
           {rippleEnabled && <RippleGroup ref={rippleRef} />}
         </LinkButton>
       );
+    case 'plain':
+      return (
+        <PlainButton as={tag} {...buttonProps}>
+          {renderContent(spinnerColor)}
+          {rippleEnabled && <RippleGroup ref={rippleRef} />}
```
Thienhuynh95, line 467: Since this `rippleEnabled` will always be disabled on `plain` intent based on this
```
    ...
    rippleEnabled: !noRipple && intent !== 'plain',
```
I think the renderRipple logic is redundant here
```suggestion
        <PlainButton as={tag} {...buttonProps}>
          {renderContent(spinnerColor)}
```

Thienhuynh95 (changes requested):

src/components/button/Button.tsx
evad1n: Agreed.

Thienhuynh95 (dismissed):
💯

Thienhuynh95 (approved):

PR #432, opened 2023-02-09
misc: make all collapse icon transitions go counter clockwise

evad1n:
## Description

This PR makes all the collapse icons rotate the same way (counter clockwise)

## Todo


Thienhuynh95 (approved):
💯

PR #433, opened 2023-02-10
fix: add button type 'button' everywhere

evad1n:
## Description

I noticed in account-web that clicking the hide password eye icon causes the submit event to fire, which is because default button type is `submit`.

This PR adds `type="button"` to all our buttons so this does not happen.

## Todo


Thienhuynh95 (approved):
💯

Thienhuynh95 (approved):

PR #434, opened 2023-02-10
refactor: remove layout effect

evad1n:
## Description

This PR removes the `useLayoutEffect` from Collapse. It originally was needed with the old design, but now it looks like it isn't needed anymore. It was breaking the SSR, so this ends up working out nicely.

## Todo


Thienhuynh95 (dismissed):
💯

Thienhuynh95 (approved):

PR #440, opened 2023-03-09
Feat/juicy button

evad1n:
## Description

This PR adds a sneaky little transition to the amino button for the extra pizzazzzzzzzz.

Also makes radio group item type generic.

## Todo


Thienhuynh95 (approved):
It's juicy 🤟

Thienhuynh95 (approved):

PR #444, opened 2023-03-27
feat(search): add fuzzy search

goodpickle:
## Description

This PR adds `fuse` to amino as our frontend fuzzy searching library.

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "3.4.33",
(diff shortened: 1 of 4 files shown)


evad1n:
I was under the impression that amino was a strictly UI library (hence the name amino-ui). Are we going to be treating this more as a our general frontend library. Like including all these utils and possibly other not strictly presentational code?

ckgardner (approved):

ckgardner (approved):

PR #451, opened 2023-04-05
refactor(CoverSheet): show error when id not found

Thienhuynh95:
## Issue

- CoverSheetAction is set to attach to an id of a `CoverSheet` wrapper that doesn't exist, so in case of the `Create Label` in the `Create Label flow` in the dashboard disappears, and it's hard to narrow down why as nothing in the dev tool console.

## Description

This PR shows the errors in the console when couldn't find the id to attach to in the whole Dom

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@amino-ui/core",
-  "version": "3.4.40",
+  "version": "3.4.41",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:amino-ui/core.git",
   "author": "Joshua Beitler <joshbeitler@gmail.com>",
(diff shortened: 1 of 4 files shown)


evad1n (dismissed):
Do you think it would be better to throw an error or just console.error?

goodpickle (approved):

Thienhuynh95:
> Do you think it would be better to throw an error or just console.error?

@evad1n I thought about that. Since this is not affecting any functionality, just not showing the button. I think it would be too much to show a 404 or 500 page 🤔 What do you think?

evad1n:
> > Do you think it would be better to throw an error or just console.error?
> 
> @evad1n I thought about that. Since this is not affecting any functionality, just not showing the button. I think it would be too much to show a 404 or 500 page 🤔 What do you think?

In the cases we have used it so far, when this occurs it makes the page unusable. I think if `CoverSheetActions` needs this to work, it should be an actual error that can't be ignored.

goodpickle (dismissed):

goodpickle (approved):

PR #453, opened 2023-04-06
Feat: Lightbox dialog for images

ckgardner:
## Jira issue

[Image lightbox](https://myzonos.atlassian.net/jira/software/projects/DOCS/boards/123?selectedIssue=DOCS-273)

## Description

I have a task on Docs that requires an image light box. Seemed like a very easy change for Amino to use the dialog in order to use all of the features and animations already built in. May be useful for other images in dashboard in the future?

## Todo


src/components/dialog/Dialog.tsx
Thienhuynh95, line 1: It looks a little bit messy here. What do you think about separating to a new Lightbox component that inherit from `_BaseDialog`? To me I feel like it would be more clear than putting new prop `image` in this component to ignore all other props.

Thienhuynh95 (changes requested):

src/components/dialog/Dialog.tsx
ckgardner: Okay! Thanks Thien!

src/components/lightbox/LightBox.tsx
```diff
@@ -0,0 +1,50 @@
...
+  width?: number;
+  noBorder?: boolean;
+  /** Close when clicking outside dialog (on the backdrop)
+   * @default true
+   */
+  closeOnBlur?: boolean;
+  /** Close on pressing 'escape' key
+   * @default true
+   */
+  closeOnEsc?: boolean;
```
Thienhuynh95, line 22:
```suggestion
  /** 
   * Close on pressing 'escape' key
   * @param closeOnEsc
   * @default true
   */
  closeOnEsc?: boolean;
```

src/components/lightbox/LightBox.tsx
```diff
@@ -0,0 +1,50 @@
...
+  className?: string;
+  children: ReactNode;
+  onClose: () => void;
+  open: boolean;
+  theme?: IAminoTheme;
+  width?: number;
+  noBorder?: boolean;
+  /** Close when clicking outside dialog (on the backdrop)
+   * @default true
+   */
```
Thienhuynh95, line 17:
```suggestion
  /** Close when clicking outside dialog (on the backdrop)
   * @param closeOnBlur
   * @default true
   */
```

Thienhuynh95 (changes requested):
Looks good to me. Just need some little adjustments to the comments.

src/components/lightbox/LightBox.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const LightBox = forwardRef<HTMLDivElement, LightBoxProps>(
+  ({
+    children,
+    className,
+    onClose,
+    open,
+    theme: _theme,
+    width,
+    noBorder = true,
+    closeOnBlur,
```
Thienhuynh95, line 37: What do you think about change the prop to have an opposite meaning?
```suggestion
    width,
    withBorder = false,
    closeOnBlur,
```
With this, if you use the component, you just need to add 1 prop `withBorder` instead of `noBorder={false}`
```
<Lightbox withBorder />
```

Thienhuynh95 (approved):
💯

PR #454, opened 2023-04-06
Fix(lightbox): Fixed forward ref issue

ckgardner:
## Description

This PR removes the forwardRef from the lightbox. The forwardRef was causing a warning in Docs that was not showing on Amino. 

## Todo


Thienhuynh95 (approved):
💯

PR #459, opened 2023-04-11
added className prop to DualCurrency

devaneereid:
## Jira issue
No Jira task

## Description
This PR adds a className to `DualCurrency` - needed for styling to render properly in dashboard.
Only want to pass down className to the DualCurrencyWrapper if it's a DualCurrency. 

## Todo


devaneereid:
Yes, I should of put DualCurrency in the description, I'll update that 👍

PR #476, opened 2023-05-01
FE-147: MDX components in Amino

ckgardner:
## Jira issue

[MDX components task](https://myzonos.atlassian.net/jira/software/projects/FE/boards/145?assignee=61f46737138863006acc2c37&selectedIssue=FE-147)

## Description

This PR moves most of the components being used on Zonos Documentation over to Amino in order to be used in other MDX projects.

## Todo


src/components/mdx-code-block/MdxCodeBlock.tsx
```diff
@@ -0,0 +1,186 @@
...
+  pre {
+    white-space: pre;
+    word-break: break-word;
+    border-radius: 0;
+    padding: 0;
+    background: transparent;
+
+    .hljs {
+      color: ${theme.gray100} !important;
+      background: transparent !important;
```
Thienhuynh95, line 53: Hmn, is this class `.hljs` using inline css? If we couldn't apply styling unless using `!important`. Would you add the commend here describe why we use !important here right here?

src/components/mdx-bread-crumbs/MdxBreadCrumbs.tsx
```diff
@@ -0,0 +1,34 @@
...
+    :hover {
+      opacity: 0.9;
+    }
+  }
+  a:first-of-type {
+    margin-left: 0;
+  }
+  span:last-of-type {
+    color: ${theme.gray300};
+  }
```
Thienhuynh95, line 25: I see that in your stories example, you are using `div` for the slash, do you think that should be span? if it's not, when will `span` scenario happen?

Thienhuynh95 (changes requested):
Looking good. Just have some question  :)

src/styles/devices.ts
```diff
@@ -0,0 +1,21 @@
...
+export const devices = {
+  mobileS: `(max-width: ${size.mobileS})`,
+  mobileM: `(max-width: ${size.mobileM})`,
+  mobileL: `(max-width: ${size.mobileL})`,
+  tabletS: `(max-width: ${size.tabletS})`,
+  tablet: `(max-width: ${size.tablet})`,
+  desktopS: `(max-width: ${size.desktopS})`,
+  desktop: `(max-width: ${size.desktop})`,
+  desktopMax: `(max-width: ${size.desktopMax})`,
+};
```
Thienhuynh95, line 21: What do you think about making these to typescript constant, so it would have literal value suggestion instead of just `string`.
```suggestion
const size = {
  mobileS: '320px',
  mobileM: '375px',
  mobileL: '525px',
  tabletS: '768px',
  tablet: '960px',
  desktopS: '1080px',
  desktop: '1300px',
  desktopMax: '1400px',
} as const;

export const devices = {
  mobileS: `(max-width: ${size.mobileS})`,
  mobileM: `(max-width: ${size.mobileM})`,
  mobileL: `(max-width: ${size.mobileL})`,
  tabletS: `(max-width: ${size.tabletS})`,
  tablet: `(max-width: ${size.tablet})`,
  desktopS: `(max-width: ${size.desktopS})`,
  desktop: `(max-width: ${size.desktop})`,
  desktopMax: `(max-width: ${size.desktopMax})`,
 as const};
```

Thienhuynh95 (changes requested):
I have one more suggestion :)
Update: This is for the review above ☝️

src/components/emoji/__stories__/Emoji.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+  display: flex;
+  width: 100%;
+  gap: 16px;
+  justify-content: stretch;
+  padding-bottom: 20px;
+  flex-wrap: wrap;
+  max-width: 100%;
+`;
+
+const StyledIcon = styled.div<{ size?: number }>`
```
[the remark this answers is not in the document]
ckgardner: Nope. Thanks!

src/components/mdx-code-block/MdxCodeBlock.tsx
ckgardner: The `.hljs` class is for the `highlight-js` library. I'm assuming we will be using the same library for our next MDX project. I had to use `!important` to override the libraries styles. I'm assuming it would need `!important` for the next project as well, but not positive 🤔 What do you think?

Thienhuynh95: Does the library use inline css for styling or just add a bunch of className to the consuming component?

src/components/mdx-bread-crumbs/MdxBreadCrumbs.tsx
ckgardner: This is how it was setup on Documentation. Maybe this will help answer your question or provide another idea?

```
<StyledBreadcrumbs className="breadcrumbs">
      <Link to="/">Home</Link> <Slash>{' /'}</Slash>
      {pathnames.map((path, index) => {
        if (path === pathToSkip) {
          return null;
        }
        return (
          <span key={path}>
            <Link to={`/${formatPath(pathnames, index)}` as IInternalPathname}>
              {formatPathName(path)}
            </Link>{' '}
            <Slash>{' /'}</Slash>
          </span>
        );
      })}
      <span>{frontmatterTitle || pathTitle}</span>
    </StyledBreadcrumbs>
```

Thienhuynh95: Yes, using the span as wrapper for the Slash (`div`), I don't think it a good HTML format 🤔  If this's the case, then I have an idea. 
What do you think about wrapping all `Link` inside of a `span`. And then using the `pseudo selector :after` for the Slash instead of using div? Then you won't have `div` inside of `span`.

Thienhuynh95 (dismissed):
💯

Thienhuynh95 (approved):

Thienhuynh95 (approved):

PR #483, opened 2023-05-12
Refactor/make select option more flexible

Thienhuynh95:
## Issue

- Select component current state only allow `options` with value `string` to be passed in, and in some scenarios, we have `options` with value is `id`, which is `number`. We need to make it to allow that.

## Description

This PR refactors to make the select `options` to allow value with `number` type and a story for that in `Select` section
image

## Todo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,6 @@ __diff_output__
 /playwright-report/
 /playwright/.cache/
 snapshots
+
+# Ignore _private_* folders (for some scratchpad scripts)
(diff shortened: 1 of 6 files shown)


goodpickle (dismissed):

evad1n:
Can you add a story with a number select?

Thienhuynh95:
> Can you add a story with a number select?

I like that idea, thanks.

evad1n (approved):

PR #490, opened 2023-05-19
FE-236: Multiple theme support

ckgardner:
## Jira issue

[Multiple theme support](https://myzonos.atlassian.net/jira/software/projects/FE/boards/145?assignee=61f46737138863006acc2c37&selectedIssue=FE-224)

## Description

This PR adds multiple theme support for Amino. The scripts used to generate some of the constants was removed. Two themes were established to start, day and night. This will set us up nicely as we introduce more themes (dawn, dusk, midnight, etc.).

Storybook is now setup to use both themes and can be changed in the toolbar of storybook. A user can switch between day, night, and side-by-side modes. 

There are a few one off color schemes that have been implemented already. The `MdxCodeBlock` and `Tooltip` are good examples of this. All React Portals were removed that would portal outside the `body` tag, as they were causing the content within to lose the `data-theme` attribute.

I removed the `DarkModeWrapper` component because we should just be able to set the `data-theme` on any component we want to force dark mode on.

## Todo


build-utils/css/README.md
```diff
@@ -31,14 +29,6 @@ This script is primarily created for generating accessible style constants and c
...
-### Overview
-Sub script `build-utils/css/buildColorConstants.ts` is primarily created for generating the color constants in folder `build-utils/css/constants/generated/colors` that can be used in any theme file (`theme.ts`). If you run `pnpm build:theme`, it will also trigger the command below.
-
-**Command**:
-```bash
-    pnpm build:colors
-```
 ### When you run the script, it will:
 
 1. Read through all the files om `build-utils/css/colorSvgs` to look for svgs that are downloaded from Figma. Base on the svg downloaded from Figma, the name of the svg will be separated into 2 parts:  "color name" and "color intensity" (Ex: "Blue500"). And the color code can be found in `fill` property of a `rect` tag in the svg. In this first step, the script will extract all of necessary data (color name, color code and color intensity) for step 2.
```
Thienhuynh95, line 34: I think this section is also related to `color` generation.
```suggestion
```

build-utils/css/README.md
```diff
@@ -202,8 +191,8 @@ Let's say we have an exported constant `testNumber` in `build-utils/csss/constan
 ## **Building process**
 
 ### Overview
-Main command for this is just `pnpm build`. This would trigger the tests for the whole build application, typescript check and eslint check, also run `pnpm build:theme` to generate css files that are based on the theme constant `theme.ts` or `_darkTheme.ts` in `build-utils/css/constants`.
-> **NOTE**: Because of running `'pnpm build:theme'` will overwrite last capture with latest content, before you run a script, run `pnpm test` first to make sure the current constant `theme.ts` and `_darkTheme.ts` in `build-utils/css/constants` doesn't have conflict with last theme capture. If there is conflict of last theme capture with current constant, resolve it either manually when you are not running `test` in `Watch mode`, or interactively by pressing `i` when you are running `test` in `Watch mode`.
+Main command for this is just `yarn build`. This would trigger the tests for the whole build application, typescript check and eslint check, also run `yarn build:theme` to generate css files that are based on the theme constant `theme.ts` or `_night.ts` in `build-utils/css/constants`.
```
Thienhuynh95, line 194: would you change `yarn` to `pnpm` in this whole README file?

build-utils/css/README.md
```diff
@@ -106,8 +96,8 @@ There would be 2 cases: the variable exists and you just want to update the colo
     - Adjust the mapping configuration for the color in the mapping array `deprecatedIntensityMapping`. In the object, the key on the left would be the legacy color suffix and the value on the right would be the new one. This mapping configuration would also add the legacy colors down below and add the `@deprecated` tag above it after the content of the new color.
     - You can also adjust the content of constant `content` in the function `generateFileContent` to modify what's being generated if needed (Like add jsdocs or change text or adding more deprecated color).
 4. Run the command below to trigger the generation process.
-    ```pnpm build:colors```
-5. Import and use the generated constant in the `theme.ts` by spreading the constant variable (you can find out more details [here](#import-generated-file-in-theme-file-themets--_darkthemets)).
+    ```yarn build:colors```
+5. Import and use the generated constant in the `theme.ts` by spreading the constant variable (you can find out more details [here](#import-generated-file-in-theme-file-themets--_nightts)).
```
evad1n, line 100: These should be `pnpm` commands, right?

Thienhuynh95 (changes requested):

src/types/ITheme.ts
```diff
@@ -0,0 +1 @@
+export type ITheme = 'day' | 'night' | 'midnight';
```
Thienhuynh95, line 1: Hmn, Amino is an external library, so the type should not be named ITheme if the types are exported. When the consumer project imports it, they would expect ITheme to be an internal type of their project. Therefore, it would be more appropriate to use a different name for the type. For external library, type should be named without prefix `I`.

src/icons/TagIcon.tsx
```diff
@@ -30,7 +30,9 @@ export const TagIcon = forwardRef<
       <defs>
         <clipPath id={`${ids[0]}`}>
           <path
-            fill={secondaryColor ? `${theme[secondaryColor]}` : '#fff'}
+            fill={
+              secondaryColor ? `${theme[secondaryColor]}` : `${theme.gray300}`
+            }
```
Thienhuynh95, line 35: This is not duotone, but the color logic of 2 colors icon is applied. Would you check if default `gray300` is applicable for this icon's secondary color, because the previous default color for this icon is `white` (#fff)?

evad1n (approved):

src/icons/TagIcon.tsx
ckgardner: Good catch on this. I rewrote the logic for `getColorVariables`. Please see that file for my changes. Thanks Thien!

Thienhuynh95 (dismissed):
Magnificient! Thanks Cade

evad1n (approved):
🥇

evad1n (approved):
🥇

PR #496, opened 2023-05-26
Fix(theme): Announcement component fixed

ckgardner:
## Description

This PR makes some unnecessary required props become optional. I also removed the type from the announcement label because I've already seen 3 different instances and there will be more to come. There is now a label and title prop.

## Todo


Thienhuynh95 (dismissed):
💯 Looks great!

goodpickle (approved):

PR #497, opened 2023-05-26
Fix(theme): Minor color changes and bug fixes

ckgardner:
## Jira issue

[Theme fixes](https://myzonos.atlassian.net/jira/software/projects/FE/boards/145?selectedIssue=FE-270)

## Description

This PR adjusts the following components/styles:
- Black transparency: uses the same hex as gray0
- Textarea: shows the placeholder now
- MdxImage: uses the shadow prop for CSS instead of an extra element
- ThemeSelect: background color fixed for toggle

## Todo

### DIFF
--- a/src/components/textarea/Textarea.tsx
+++ b/src/components/textarea/Textarea.tsx
@@ -75,10 +75,7 @@ const StyledTextarea = styled.textarea<TextareaType>`
     padding: ${theme.space24} 4px ${theme.space8} ${theme.space16};
 
     ::placeholder {
-      transition: ${theme.transition};
       opacity: 0;
-      color: ${theme.gray400};
-      font-weight: 400;
     }
 
     &.has-content,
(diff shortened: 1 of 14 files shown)


src/components/textarea/Textarea.tsx
```diff
@@ -74,6 +74,13 @@ const StyledTextarea = styled.textarea<TextareaType>`
   &.has-label {
     padding: ${theme.space24} 4px ${theme.space8} ${theme.space16};
 
+    ::placeholder {
+      transition: ${theme.transition};
+      opacity: 0;
+      color: ${theme.gray400};
+      font-weight: 400;
+    }
+
```
Thienhuynh95, line 83: Now I'm thinking more about this. What do you think about having `::placeholder` outside the scope of `&.has-label` with css like `color`, `transition` and `font-weight`. And here you would have just `opaicity: 0`? So that in the regular `textarea` with no `label`, the placeholder style would stay the same.
```suggestion
    ::placeholder {
      opacity: 0;
    }

```

Thienhuynh95 (changes requested):

Thienhuynh95 (approved):

PR #503, opened 2023-06-07
stories: mdx doc stories for theme usage

evad1n:
## Description

This PR adds a doc story for theme usage.

~No version bump as these are internal story only changes.~ Hook has an added optional prop to modify the root.

Link: https://amino-git-feat-theme-usage-docs-zonos.vercel.app/?path=/docs/amino-themeselect-usage--docs

I went back and forth on the doc (mdx) story loading, but I figured it would be too specific and rare to have custom logic for finding those stories. As a result, if you want to add a doc story, then you will have to manually set the title.

### DIFF
--- a/.storybook/buildStories.ts
+++ b/.storybook/buildStories.ts
@@ -1,3 +1,4 @@
+import { glob } from 'glob';
 import { capitalize } from '../build-utils/css/utils/capitalize';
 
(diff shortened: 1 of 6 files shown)


Thienhuynh95 (approved):
["L", "O", "V", "E", "L", "Y"].join('')

Thienhuynh95 (approved):

PR #504, opened 2023-06-08
Code tag

nick-lanners:
## Jira issue

[FE-416 - Add Amino code tags component](https://myzonos.atlassian.net/browse/FE-146)

## Description

This PR adjusts the `Tag` component to produce a `code` variant create stories for new component form

## Todo

### DIFF
--- a/src/components/tag/Tag.tsx
+++ b/src/components/tag/Tag.tsx
@@ -1,3 +1,4 @@
+import { unset } from 'lodash';
 import type { ReactNode } from 'react';
 
 import { RemoveIcon } from 'src/icons/RemoveIcon';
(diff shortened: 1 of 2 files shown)


evad1n:
Code looks good and matches the design, but I left some comments on the Figma design as this does not fit with our current Tag design. It adds a disabled state which we don't implement in our current Tags and then doesn't have any styling for hover, active states. Waiting for designer feedback.

nick-lanners:
> Code looks good and matches the design, but I left some comments on the Figma design as this does not fit with our current Tag design. It adds a disabled state which we don't implement in our current Tags and then doesn't have any styling for hover, active states. Waiting for designer feedback.

Yeah I was wondering about that. My initial thought was to remake the `Tag` component with all of the new styles in the Figma, but was advised to just alter the current to make work for the `Code Tag`. I am happy to make changes based on the designer feedback :)

evad1n:
@goodpickle https://www.figma.com/file/WnKnmG7L3Q74hqPsw4rbEE/Amino-2.0?type=design&node-id=5114-93930&mode=design#466757432

PR #505, opened 2023-06-08
fix(GraphMatrix): workaround for cursor jumping

nick-lanners:
## Jira issue

[FE-221 - Amino: GraphMatrix - Graphiql cursor jump when typing fast](https://myzonos.atlassian.net/browse/FE-221)

## Description

This PR implements a workaround to prevent the cursor from jumping to the beginning of the query window when typing quickly
this fix comes from:
https://github.com/graphql/graphiql/issues/2368#issuecomment-1176119332 possibly doesn't work all the time, GitHub issue is not being worked on as GraphiQL is looking to reimplement the Explorer plugin soon

## Todo


evad1n (approved):

goodpickle:
@nick-lanners It looks like this needed a bump before I merged it. Will you run `yarn version:patch` on another branch, so we can get this patched?

PR #507, opened 2023-06-08
eslint rules to order props

nick-lanners:
## Jira issue
[FE-303 - Apply `eslint` rule to sort all props in all projects](https://myzonos.atlassian.net/browse/FE-303)

## Description
This PR adds some eslint rules to auto-fix and sorting all props alphabetically 🎉 
Added eslint rules:
- typescript-custom-sort-keys/interface [ref](https://github.com/prash471/eslint-plugin-typescript-custom-sort-keys)
  - sorting props `type` and `interface`
- sort-keys/sort-keys-fix [ref](https://github.com/namnm/eslint-plugin-sort-keys)
  - sorting object keys (comment is tied with the key, so the comment will also be sortd out, this rule doesn't sort destructure object)
- sort-destructure-keys/sort-destructure-keys [ref](https://github.com/mthadley/eslint-plugin-sort-destructure-keys)
  - sorting destructure object
- react/jsx-sort-props [ref](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md)
  - sorting react props (key will be moved to first possition)

## Todo


.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     'prettier',
     'plugin:storybook/recommended',
   ],
+  ignorePatterns: ['**/generated/**'],
```
evad1n, line 17: We shouldn't need this as we have no generated files in amino.

.eslintrc.js
```diff
@@ -39,8 +43,48 @@ module.exports = {
...
+       * Turn off sort-keys for generated files (it will take longer to sort and there is no need to sort the generated files)
+       */
+      files: ['**/generated/**/*.ts'],
+      rules: {
+        'sort-destructure-keys/sort-destructure-keys': 'off',
+        'sort-keys': 'off',
+        'sort-keys/sort-keys-fix': 'off',
+        'typescript-custom-sort-keys/interface': 'off',
+      },
+    },
```
evad1n, line 57: We shouldn't need this as we have no generated files in amino. (2)

.eslintrc.js
```diff
@@ -39,8 +42,48 @@ module.exports = {
...
+       * Turn off sort-keys for generated constants file (we don't want to sort because of comments, but want to do other linting)
+       */
+      files: ['src/styles/constants/*.ts'],
+      rules: {
+        'sort-destructure-keys/sort-destructure-keys': 'off',
+        'sort-keys': 'off',
+        'sort-keys/sort-keys-fix': 'off',
+        'typescript-custom-sort-keys/interface': 'off',
+      },
+    }
```
Thienhuynh95, line 56: Just for record. We shouldn't change the order of generated constant file, because it will make this will mess up with the section comments.

Thienhuynh95 (approved):
Looking good!

evad1n (approved):

PR #523, opened 2023-06-28
refactor: remove emoji

evad1n:
## Description

This PR removes the Emoji component from amino. It was tied closely with the mdx, and had external dependencies on the consumer project having a specific structure. It is being moved to docs.

Changes `useAminoTheme` hook to use `documentElement` as root instead of body, when root is true.

## Todo


src/components/emoji/Emoji.tsx
```diff
@@ -1,26 +0,0 @@
...
-};
-
-export const Emoji = ({ code }: Props) => {
-  const emojiFileName = emojiFileNames.find(({ emoji }) => code === emoji);
-
-  if (emojiFileName) {
-    const attribution = `Attribution for Twitter Emojis: ©️ Twitter, Inc.`;
-    const alt = `${emojiFileName.emoji} ${attribution}`;
-    return <StyledImage alt={alt} src={`/emojis/${emojiFileName.fileName}`} />;
-  }
```
Thienhuynh95, line 22: It looks like we have folder `emoji` here. Do we also need to remove this folder too? 🤔

Thienhuynh95 (changes requested):

src/components/emoji/Emoji.tsx
evad1n: No, it is actually an external reference (for the consuming project), hence why it didn't make any sense  to be here.

Thienhuynh95 (approved):

Thienhuynh95 (approved):

PR #526, opened 2023-07-07
Feat: hoverBackground prop for buttons

ckgardner:
## Description

This PR adds a prop for custom hover color for special situations. May need to add a hoverColor prop for text color later, but not needed for what I'm doing.

## Todo

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -122,7 +122,7 @@ const Primary = styled(AminoButton)`
 
   &:not([disabled]) {
     &:hover {
-      background: ${p => getAminoColor(p.hoverColor) || theme.blue400};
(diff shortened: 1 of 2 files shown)


src/components/button/Button.tsx
```diff
@@ -318,6 +319,7 @@ type ButtonBase = {
   className?: string;
   color?: Color | 'inherit';
   disabled?: boolean;
+  hoverBackground?: Color;
```
evad1n, line 322: I forget why we allowed 'inherit' as an option, but it might be a good idea to add it for the hoverBackground as well? I honestly don't know, but it couldn't hurt

evad1n (dismissed):

Thienhuynh95 (approved):

PR #530, opened 2023-07-13
Refactor/improvement

Thienhuynh95:
## Jira issue

[[Zonos Elements] initialize the project with all of the bundling process from Amino](https://myzonos.atlassian.net/browse/FE-402)

## Description

This PR:
- [x] upgrades most of the libraries to the latest
  - prettier to the latest (3.0). Jest is not happy with prettier 3.0 esm at all, but `vitest` does. So I added `vitest` and replace it with `jest` in most places except `storyshots`. (https://prettier.io/blog/2023/07/05/3.0.0.html)
  - replace `jest` with `vitest`.
  - upgrade `react-data-grid` to the latest (https://github.com/adazzle/react-data-grid/pull/3239) 
    - update prop/varaible name changes
  - ~~upgrade `@graphiql/plugin-explorer`, `@graphiql/react` and `graphiql` to the latest~~. I tried to upgrade and it looks like it's having a runtime issue when consuming from another project with the error `Tooltip must be place inside of "TooltipProvider"` from `radix-ui/tooltip`(which is being used by `graphiql`). It's weird that copying the `GraphMatrix` from amino over and it has no issue if using the component, but if use the one from the bundled amino version, it's just error out. This breaking change upgrade of graphiql would possibly resolve this issue https://myzonos.atlassian.net/browse/FE-221. But skip it for now until I have more bandwidth to figure that out 😭 
- [x] improve eslint rules
  - fix the import sorting group
  - extends vitest recommended plugin
  - set `trailing-comma` to `all` as `prettier` 3.0 default their config to `all` and it would be better for cleanner git diff when reviewing
- [x] improve building process
  - close the bundle connection when finish bundling as rollup recommened.
- [x] apply `lint --fix` with the new eslint rules
- [x] move the `GridSpacing` to `types` folder since it would make more sense to be there and the build script will not need to warn about `generate empty chunk` when bundling.

Ps: this has changed the location of `GridSpacing`, so bump version is needed.
## Todo


evad1n:
How about react 18? 🥺

goodpickle (dismissed):

Thienhuynh95:
> How about react 18? 🥺

It's on my list. But get this PR out first, it has quite big file changes, and more important this upgrade works locally for `graph-explorer` and `dashboard` 😅

goodpickle (approved):

PR #539, opened 2023-07-24
Fix/build transpile paths

evad1n:
## Description

This PR fix an issue with the type declarations and svg loading with absolute paths.

Also improves the storybook theming:
- amino theme and storybook theme now sync
- side by side is iframed so portals work together

Lints storybook files now

~There is still in issue at the theme select side by side. (`/?path=/story/amino-themeselect--cards&globals=theme:side-by-side`). It uses local storage, which goes through iframes so...~

I hard coded a fix to the above story.

## Todo


.eslintrc.js
```diff
@@ -53,6 +53,12 @@ module.exports = {
         'import/no-extraneous-dependencies': 'off',
       },
     },
+    {
+      files: ['.storybook/**'],
+      rules: {
+        'no-relative-import-paths/no-relative-import-paths': 'off',
+      },
+    },
```
evad1n, line 61: Storybook configs that define their own aliases understandably don't work.

src/components/dialog/_BaseDialog.tsx
```diff
@@ -60,7 +59,6 @@ export const BaseDialog = ({
   themeOverride,
   width,
 }: BaseDialogProps) => {
-  const { aminoTheme } = useAminoTheme();
```
evad1n, line 63: This shouldn't be necessary anymore. I tested in dashboard and it works fine. Maybe there is an edge case I missed though. @ckgardner

.storybook/preview.tsx
```diff
@@ -20,36 +27,82 @@ const ThemeBlock = styled.div<{
...
+  const { aminoTheme, setAminoTheme } = useAminoTheme({ root: true });
+  const previousStorybookTheme = usePrevious(storybookTheme);
+
+  const sameTheme = storybookTheme === aminoTheme;
+  const storybookChanged = previousStorybookTheme !== storybookTheme;
+
+  // Sync amino theme with storybook
+  useEffect(() => {
+    if (previousStorybookTheme !== storybookTheme) {
+      if (!sameTheme && storybookTheme !== 'side-by-side') {
```
goodpickle, line 56: What do you think about avoiding nested if statements?
```suggestion
    if (previousStorybookTheme !== storybookTheme && !sameTheme && storybookTheme !== 'side-by-side') {
```

goodpickle (changes requested):

.storybook/preview.tsx
evad1n: Leftover from my brain trying to figure out the logic to sync these states.. 🤦

build-utils/build.ts
```diff
@@ -55,7 +55,7 @@ const bundlePackage = async (
         resolveOnly: [''],
       }),
       image({
-        limit: 10000,
+        // limit: 10000,
```
ckgardner, line 58: this needed?

src/components/dialog/_BaseDialog.tsx
ckgardner: `|| aminoTheme` was used because the dialog is rendered outside the parent `data-theme` attribute. If you fixed that, then this shouldn't be necessary.

build-utils/build.ts
evad1n: for fun

src/components/dialog/_BaseDialog.tsx
evad1n: Yeh the new root hook sets it to the `documentElement` which is impossible(?) to render outside of I believe...

Thienhuynh95 (dismissed):
Magnificent!

src/utils/hooks/usePrevious.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from 'react';
+
+export const usePrevious = <T extends unknown>(value: T): T | undefined => {
+  const ref = useRef<T>();
+  useEffect(() => {
+    ref.current = value;
+  });
```
Thienhuynh95, line 7: Does this need dependencies? 🤔

goodpickle (changes requested):

goodpickle (dismissed):

goodpickle (approved):

PR #542, opened 2023-07-25
New-buttons

nick-lanners:
## Jira issue

[FE-9 - New Amino button styles](https://myzonos.atlassian.net/browse/FE-9)

## Description

This PR:
- update buttons to match the Amino 2.0 styles:
https://www.figma.com/file/WnKnmG7L3Q74hqPsw4rbEE/Amino-2.0?node-id=1311%3A20642&mode=dev
- move Button to LegacyButton so as not to break everything
- update test snapshots
- update theme variables to include Amino 2.0 variables

PREVIEW: https://amino-git-new-buttons-zonos.vercel.app/?path=/story/amino-button--default

## Todo


src/components/button/Button.tsx
```diff
@@ -112,207 +117,239 @@ const AminoButton = styled.button<ButtonProps<GroupTag>>`
...
-  background: ${p => getAminoColor(p.background) || theme.primary};
-  color: ${p => getAminoColor(p.color) || theme.gray0};
+  background-color: ${p =>
+    getAminoColor(p.background) || p.outline ? theme.gray0 : theme.primary};
+  color: ${p =>
+    getAminoColor(p.color) || p.outline ? theme.primary : theme.gray0};
+  box-shadow: ${p => (p.outline ? 'none' : theme.shadowButtonPrimary)};
+  ${p =>
+    p.outline
+      ? 'border: 1px solid var(--amino-blue-300); border-radius: var(--amino-radius-6);'
```
Thienhuynh95, line 135:
```suggestion
      ? `border: 1px solid ${theme.blue300}; border-radius: ${theme.radius6};`
```

src/components/button/Button.tsx
```diff
@@ -112,207 +117,239 @@ const AminoButton = styled.button<ButtonProps<GroupTag>>`
...
+
+  &[disabled] {
+    .content {
+      opacity: 0.6;
+    }
   }
 `;
 
-type IntentProps = 'outline' | 'subtle' | 'text' | 'link' | 'plain' | Intent;
+type IntentProps = 'subtle' | 'link' | 'plain' | Intent;
```
evad1n, line 352: If we are using this opportunity to complete change the new button API, then I would prefer this prop be called `variant`

src/components/button/Button.tsx
```diff
@@ -112,207 +117,239 @@ const AminoButton = styled.button<ButtonProps<GroupTag>>`
...
   }
 `;
 
-const Danger = styled(AminoButton)`
-  background: ${p => getAminoColor(p.background) || theme.red600};
-  color: ${p => getAminoColor(p.color) || theme.gray0};
+const Standard = styled(AminoButton)`
+  color: ${p => getAminoColor(p.color) || theme.textColor};
+  background: ${p => getAminoColor(p.background) || theme.surfaceColor};
+  border: 1px solid ${p => (p.outline ? theme.gray300 : `${theme.gray600}1F`)};
```
Thienhuynh95, line 205: I don't think adding the letter after the css variable will work 🤔 Does this work?

ckgardner (commented):
I've been playing around on the preview url and see a couple minor issues. May be good to get Connor or Claire to QA the preview branch. For instance, the out hover on the buttons has a pretty bad flicker.

Thienhuynh95 (changes requested):

src/components/button/Button.tsx
evad1n, line 205: This `1F` here is invalid. You can't really add alpha values to CSS variables directly. Probably needs a new variable.

src/types/Intent.ts
```diff
@@ -1,6 +1,7 @@
 export type Intent =
   | 'default'
   | 'primary'
+  | 'standard'
   | 'secondary'
   | 'info'
   | 'success'
```
evad1n, line 7: Here as well. We can keep the old `Intent` type because it used in other components, but I think making a new type called `Variant` is better moving forward.

src/components/button/Button.tsx
nick-lanners: It does! I did a bunch of playing around with it. It was a really interesting discovery.

nick-lanners: It does! But only in certain circumstances. The color variables, for example, are just the hex code string. This one, `gray600` is the same as writing `'#6A6D77'` and because I'm dropping that in a string, following it up with `1F` creates `'#6A6D771F'` which add the transparency to the border. I was pretty happy to discover that. We can see a similar effect in the `Spinner.tsx` component with the border sizes: following up a string interpolation with more text (in that case, the `px` after a calculation based on the props).

nick-lanners:
> I've been playing around on the preview url and see a couple minor issues. May be good to get Connor or Claire to QA the preview branch. For instance, the out hover on the buttons has a pretty bad flicker.

Oh shoot. I had put in a solution to that early but changed it, forgetting why I did it the way I did. I can fix that.

src/components/button/Button.tsx
```diff
@@ -112,207 +117,206 @@ const AminoButton = styled.button<ButtonProps<GroupTag>>`
...
```
ckgardner, line 192: Not sure if you are still wanting a PR review. I saw it in slack and came to review and saw this was still an issue. If you are still working on it, what do you think about making it a draft PR?

ckgardner (changes requested):

ckgardner:
> > I've been playing around on the preview url and see a couple minor issues. May be good to get Connor or Claire to QA the preview branch. For instance, the out hover on the buttons has a pretty bad flicker.
> 
> Oh shoot. I had put in a solution to that early but changed it, forgetting why I did it the way I did. I can fix that.

Hover looks much better. Nice work!

src/components/button/Button.tsx
nick-lanners: Just fixed that bit. Lunch interrupted my work. This is ready for another review as of 2:05pm, July 26th.

src/components/button/__stories__/Button.stories.tsx
```diff
@@ -174,53 +177,88 @@ const Template: StoryFn<ButtonProps> = props => (
...
-export const TextButton = Template.bind({});
-TextButton.args = {
-  children: 'Text Button',
-  intent: 'text',
-};
+// export const TextButton = Template.bind({});
+// TextButton.args = {
+//   children: 'Text Button',
+//   intent: 'text',
+// };
```
Thienhuynh95, line 258:
```suggestion
```

src/components/button/Button.tsx
```diff
@@ -387,15 +408,16 @@ export function Button<T extends GroupTag = 'button'>({
...
   const tag = _tag || 'button';
   const renderContent = (_spinnerColor?: SpinnerProps['color']) => (
     <>
-      {!iconRight && icon}
-      {children}
-      {iconRight && icon}
-      {intent !== 'plain' && loading && (
+      <span className="content">{!iconRight && icon}</span>
+      <span className="content">{children}</span>
+      <span className="content">{iconRight && icon}</span>
```
Thienhuynh95, line 419: Could you try to change to another tag for these? `span` is an inline element, and `children` is possibly a `block` element. And having a block inside a inline is an invalid html standard 🤔 
```suggestion
      <span className="content">{!iconRight && icon}</span>
      <div className="content">{children}</div>
      <span className="content">{iconRight && icon}</span>
```

Thienhuynh95 (changes requested):

src/components/button/Button.tsx
```diff
@@ -348,12 +367,15 @@ type ButtonBase = {
...
    */
   noRipple?: boolean;
+  outline?: boolean;
   size?: Size;
   /** Color for the spinner when in a loading state */
   spinnerColor?: SpinnerProps['color'];
   tabIndex?: number;
   themeOverride?: Theme;
   type?: 'button' | 'reset' | 'submit';
+  /** @default 'secondary' */
```
evad1n, line 377:
```suggestion
  /** @default 'standard' */
```

src/components/button/__stories__/Button.stories.tsx
```diff
@@ -59,9 +59,9 @@ const ButtonRowLabel = styled(Text)`
 
 const ButtonRow = ({
   disabled,
-  intent,
   label,
   loading,
+  variant: intent,
```
evad1n, line 64:
```suggestion
  variant,
```

Thienhuynh95 (dismissed):

evad1n:
@ckgardner your approval is needed

ckgardner (approved):

PR #545, opened 2023-07-27
Feat/export base dialog

evad1n:
## Description

This PR refactors and exports the `BaseDialog` component.

Any components that can use it will now use it. I added props to support all the use cases I could see.

The main difference is instead of another backdrop element existing for the opacity fade, it now animates the background color transparency so we don't need another element to get around the opacity and children interaction.

This will consolidate some of the gnarly logic we have reimplemented in dashboard and other upstream projects.

Also:
- adds an amino favicon to our storybook.
- upgrades storybook to latest (no changes were needed)

## Todo


src/components/lightbox/LightBox.tsx
```diff
@@ -1,50 +1,12 @@
...
-    data-theme={themeOverride}
-    noBorder={!withBorder}
-    onClose={onClose}
-    open={open}
-    width={width}
-  >
-    {children}
-  </BaseDialog>
+export const LightBox = ({ children, ...props }: LightBoxProps) => (
+  <BaseDialog {...props}>{children}</BaseDialog>
```
evad1n, line 11: This component is now actually identical to the BaseDialog component. The only difference is the `onClose` prop is required instead of optional.

src/components/dialog/BaseDialog.tsx
```diff
@@ -0,0 +1,181 @@
...
+  width?: number;
+  /**
+   * Disable the backdrop color
+   * @default true
+   */
+  withBackdrop?: boolean;
+  /**
+   * @default true
+   */
+  withBorder?: boolean;
```
evad1n, line 78: This naming convention for a border prop was used elsewhere, so I conformed.

goodpickle (dismissed):

goodpickle (approved):

PR #549, opened 2023-08-01
feat: use mui collapse for collapse

evad1n:
## Description

This PR swaps out the Collapse implementation with material-ui's Collapse because theirs is 🔥 . Tweaked the Collapse stories to more accurately reflect our use cases.

Originally needed fixing because it did not behave well when nested.

Adds some licenses as an afterthought.

## Todo


ckgardner (dismissed):
exciting!

ckgardner (approved):

PR #551, opened 2023-08-08
feat: customized toast positioning

nick-lanners:
## Jira issue
[FE392 - Classify: Amino: Error toast alignment](https://myzonos.atlassian.net/browse/FE-392)

## Description

This PR does the following: 

add ability to determine what direction the toast animates in from:
- left, right, top, bottom

add ability to determine where the toast appears on the page.
Usage:
- `const notify = useNotify()`
- `notify(<toast message>, {direction, duration, intent}, {bottom?: <string>, left?: <string>})`
- `{bottom, left}` location object is separate and optional to not muddy up the `Toast` component
- location object only used by `ToastContext` so don't want to pass in `props` object

updated stories accordingly
- add Direction, Horizontal Offset, Vertical Offset to Custom Toast panel

## Todo

### DIFF
--- a/src/components/toast/__stories__/ToastConsumer.tsx
+++ b/src/components/toast/__stories__/ToastConsumer.tsx
@@ -4,7 +4,9 @@ import styled from 'styled-components';
 
 import { LegacyButton } from 'src/components/button/LegacyButton';
 import { Input } from 'src/components/input/Input';
+import { Select } from 'src/components/select/Select';
 import { VStack } from 'src/components/stack/VStack';
+import type { Direction } from 'src/components/toast/Toast';
 import { ToastContext } from 'src/components/toast/ToastContext';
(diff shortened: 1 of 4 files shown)


src/components/toast/__stories__/ToastConsumer.tsx
```diff
@@ -83,8 +91,48 @@ export const ToastConsumer = () => {
             rows={5}
             value={message}
           />
+          <Select
+            label="Direction"
+            onChange={v => {
+              if (v) {
+                setDir(v);
+              }
+            }}
```
ckgardner, line 100:
```suggestion
            onChange={v => v && setDir(v)}
```
What do you think about this shorthand?

ckgardner (changes requested):
This will be a sweet feature 🔥

ckgardner (approved):

PR #555, opened 2023-08-30
Feat: SimpleTable and Filters

evad1n:
## Description

This PR adds new components.

- SimpleTable - Basic table inspired by other frameworks table designs. In the future we could have a more complicated table to handle collapses/expansions that are present in a lot of designs.
- Filters (current API is basic, as more use cases arrive we might be able to add more logic here in amino to handle complex things like dates).

These components arose organically through implementation in admin and then in dashboard, and then back again in admin. After 4 passes of refinement/usage, they should be stable enough to use in most cases.

`dayjs` is now a peer-dep for the dates.
Adds the deep reducer and related types.

Fixes:
- HTML label violations (spaces in id)
- storybook stories are all full height (see the current menu button story for the problem)

Misc:
- Upgrade typescript to latest
- refactors the country URLs hooks used in storybook to consolidate code

## Todo


src/components/filter/filter-amount/FilterAmount.tsx
```diff
@@ -0,0 +1,156 @@
...
+      case 'greater':
+        dispatchAmounts({ amount1: editingAmount1, amount2: null });
+        setNewFilterText(`Greater than ${amount1String}`);
+        break;
+      case 'less':
+        dispatchAmounts({ amount1: null, amount2: editingAmount1 });
+        setNewFilterText(`Less than ${amount1String}`);
+        break;
+      default:
+        // Handle default case if necessary
```
ckgardner, line 96:
```suggestion
```
I would think this is self explanatory 🤷🏼‍♂️ Sorry if you copied this from me lol

src/components/filter/filter-date/temp.tsx
ckgardner, line 1: What's this `temp` file?

ckgardner (changes requested):
These components in Amino is an awesome upgrade.

src/components/filter/filter-date/temp.tsx
evad1n: oops I need to rename that file. It was to fix the import cycle but then I didn't rename it...

src/components/checkbox/Checkbox.tsx
```diff
@@ -122,10 +124,10 @@ export const Checkbox = ({
       checked={checked}
       className={['amino-input-wrapper', disabled ? 'disabled' : ''].join(' ')}
       data-testid={testId}
-      htmlFor={label}
+      htmlFor={labelAsHtmlAttribute}
```
Thienhuynh95, line 127: Maybe you can just use `testId` variable instead 🤔

Thienhuynh95 (changes requested):

src/components/checkbox/Checkbox.tsx
evad1n: I think if we are using labels in forms, that id is more semantically correct. It would be more functionally correct as well, but we wrap the switch/checkbox in the whole label anyway so it doesn't make a difference.

ckgardner (approved):

src/components/simple-table/SimpleTable.tsx
Thienhuynh95, line 1: Looks good. But now we have 2 ways of making a table in amino. One is using composed `<Table>, <TableHead>, <TableCell>, ...` and the other is using a single component and passing props in to render. I don't know if having two ways of rendering a table is what we want 🤔. I think either this component would extend and use the existing component so it would share somewhat the same style or just keep either one of them, not both. Because if the table design change, we now need to go and update for both components and may just update one and forget doing the other.
@evad1n What do you think? 🤔

src/components/checkbox/Checkbox.tsx
Thienhuynh95: Sounds good 😉

src/components/simple-table/SimpleTable.tsx
evad1n: This table was created to simplify the need to manually created a table using the existing amino components. It was designed to fit in with newer designs. I don't think it is necessary to force it to use existing Table components. 

For more complicated tables (expandable), they will need to be manually created using the Amino components.

So I guess I am saying the existing Table components are there for more fine grained control, while the SimpleTable is more opinionated to be... simpler.

@ckgardner thoughts?

evad1n: In the future I think the simple table will be what we are using, and then we can add more advanced functionality (expansion) and phase out the old table.

Thienhuynh95: Awesome, I just pointed it out so that we can have some ideas and agree on how to use the Amino table components or this SimpleTable. And they sounds great.

Thienhuynh95 (approved):

src/components/simple-table/SimpleTable.tsx
ckgardner: I personally have loved the simple table. Very simple implementation for what would typically be more complex. The resolution of the conversation sounds good to me.

ckgardner (approved):

PR #556, opened 2023-09-01
Fix/collapse table row

Thienhuynh95:
## Issue

Sometimes the style that set height for the `TableRow` is put below the style that set the height to 0 when nextjs build, so the height is not set properly, cause some ui issue.
image

## Description

This PR add triple target to the class itself to makes it so that height is set properly when collapse.
image

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.1.8",
+  "version": "5.1.9",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:Zonos/amino.git",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


evad1n:
What do you mean by "sometimes"? I have experienced this issue as well, due to the stylesheet ordering by styled components. I fixed my issue by using `&&` upstream. I am hesitant to use the super selector here in amino as it is very much like using `!important`. It seems like we could handle this with a class added when it is collapsed to be more specific.

Thienhuynh95:
It doesn't happen with me locally, but it appears to happen in prod. Well, super selector like &&& is not like !important though. At least inline style can still affect on the selector, not like !important.

goodpickle (approved):

PR #557, opened 2023-09-05
Fix: button width default to fit-content

ckgardner:
## Description

This PR adds a prop to defined the button width to `fit-content`. I've run in to multiple scenarios where if a button is in a grid (HStack or VStack), it automatically takes the full width. 

## Todo


Thienhuynh95:
Can you write a stories that demonstrates the scenario you mentioned?

src/components/button/Button.tsx
```diff
@@ -89,6 +89,7 @@ const AminoButton = styled.button<ButtonProps<GroupTag>>`
   cursor: pointer;
   white-space: nowrap;
   padding: ${p => getPadding(p.size)};
+  width: fit-content;
```
Thienhuynh95, line 92: The button currently is in "display:flex", so obviously it will need to have a wrapper or a flex parent to handle the width. If you want it to fit the button content. What do you think about use `inline-flex` instead 🤔 ?

Thienhuynh95 (changes requested):

src/components/button/Button.tsx
ckgardner: It has nothing to do with flex from what I can see. I changed it to `block` and it didn't change the width.

src/components/button/__stories__/Button.stories.tsx
```diff
@@ -262,3 +282,6 @@ PlainButton.args = {
   children: 'Plain button',
   variant: 'plain',
 };
+
+const BadTemplate: StoryFn<ButtonProps> = () => <Button>Test</Button>;
+export const Test = BadTemplate.bind({});
```
Thienhuynh95, line 287: What is this bad template for 🤔

Thienhuynh95 (approved):

PR #558, opened 2023-09-07
updated styling on tags and badges

connorvang:
## Description

Adjusts the styling for the tags and badges components to match Figma

### DIFF
--- a/src/components/badge/Badge.tsx
+++ b/src/components/badge/Badge.tsx
@@ -49,7 +49,7 @@ const StyledBadge = styled.div<StyledProps<StyledBadgeProps>>`
 
   // size
   &.small {
-    padding: 2px 4px;
+    padding: 2px ${theme.space4};
     border-radius: ${theme.radius4};
     gap: 2px;
   }
--- a/src/components/badge/Badge.tsx
+++ b/src/components/badge/Badge.tsx
@@ -18,7 +18,7 @@ type StyledBadgeProps = {
 
 const StyledBadge = styled.div<StyledProps<StyledBadgeProps>>`
   display: flex;
-  gap: ${theme.space8};
+  gap: ${theme.space4};
   font-size: ${theme.fontSizeS};
   padding: 4px 8px;
   text-align: center;
@@ -34,6 +34,8 @@ const StyledBadge = styled.div<StyledProps<StyledBadgeProps>>`
   }
   svg {
     order: ${p => (p.$iconRight ? '2' : '')};
+    height: 16px;
+    width: 16px;
   }
 
   &.bold {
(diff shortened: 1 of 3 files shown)


Thienhuynh95:
@connorvang It looks like the lint is error out.
Will you run `pnpm lint:prod --fix` and re-commit changes to resolve this?

src/components/tag/Tag.tsx
```diff
@@ -20,25 +20,24 @@ export interface TagProps {
...
-  background-color: ${theme.gray200};
+  font-size: ${theme.fontSizeS};
+  background-color: ${theme.gray100};
   border-radius: ${theme.radius6} 0 0 ${theme.radius6};
-  gap: ${theme.space8};
   font-weight: normal;
-  padding: 0 ${theme.space8};
-
-  color: ${theme.gray800};
+  padding: ${p => (p.size === 'base' ? '2px 4px' : '4px 4px 4px 8px')};
```
Thienhuynh95, line 35: If we don't have the variable exist, either to round it up or down so we can utilize the amino variable. Or if it looks really bad, then use the number `px` that you think would look good. It's best to try to use the amino variables that are available, then we would have a unified styling throughout the app.
```suggestion
  padding: ${p => (p.size === 'base' ? `2px ${theme.space4} : `${theme.space4} ${theme.space4} ${theme.space4} ${theme.space8}`)};
```

src/components/badge/Badge.tsx
```diff
@@ -47,7 +49,9 @@ const StyledBadge = styled.div<StyledProps<StyledBadgeProps>>`
 
   // size
   &.small {
-    padding: 2px 8px;
+    padding: 2px 4px;
```
Thienhuynh95, line 52:
```suggestion
    padding: 2px ${theme.space4};
```

Thienhuynh95 (changes requested):

src/components/tag/Tag.tsx
connorvang: Ill get that updated. I noticed we don't have a 2px variable in Amino yet and I was trying to use it quite a bit. I think thats the only variable we don't have that we use in our designs since we use a 4px grid everywhere and 2px when needed for smaller spacing.

Thienhuynh95: That was our original plan for spacing. We might consider using 2px grid if this appears to be used more often in the future.

Thienhuynh95 (dismissed):

Thienhuynh95:
@connorvang You now can bump the version by running this command
```
pnpm version:patch
```

Thienhuynh95 (approved):

PR #561, opened 2023-09-12
Updated the switch, checkbox, and radios

connorvang:
## Description

Updates the switch component to match the Figma stylings.

Fixes to radios and checkboxes to show the glow/shadow when active.


evad1n (dismissed):
you'll have to pull from main before updating the version

evad1n (approved):

PR #562, opened 2023-09-13
Refactor/storage hooks

evad1n:
## Description

This PR refactors the `useStorage` hook to use new react hooks instead of swr. swr was really just an unnecessary layer which added extra latency to a native browser capability.

Also fixes a bug where if `json` was true but there was no `schema`, then it would not return the parsed value.

Should also be SSR safe. There is support inside the hook https://react.dev/reference/react/useSyncExternalStore#adding-support-for-server-rendering

However, the snapshot value does shallow comparison (doesn't work for objects), therefore I opted to just use it for render comparison, and get the parsed value separately anyways. The getter is server safe as well and defaults to the `defaultValue`

EDIT: disregard the above, it appears `getServerSnapshot` is required regardless...

- Filters now get applied when pressing enter

## Todo

### DIFF
--- a/src/utils/hooks/__tests__/useStorage.test.ts
+++ b/src/utils/hooks/__tests__/useStorage.test.ts
@@ -6,6 +6,7 @@ import { useStorage } from 'src/utils/hooks/useStorage';
 describe('useStorage', () => {
   afterEach(() => {
     localStorage.clear();
+    sessionStorage.clear();
   });
 
   test('should set an array of strings', async () => {
@@ -53,7 +54,7 @@ describe('useStorage', () => {
         json: true,
         key: 'people-array',
         schema: person,
-        type: 'local',
+        type: 'session',
       }),
     );
 
--- a/src/utils/hooks/__tests__/useStorage.test.ts
+++ b/src/utils/hooks/__tests__/useStorage.test.ts
@@ -13,7 +13,6 @@ describe('useStorage', () => {
     const { result } = renderHook(() =>
       useStorage({
         defaultValue: [],
-        json: true,
         key: 'people-array',
         schema: z.array(z.string()),
         type: 'local',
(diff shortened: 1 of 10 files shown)


src/utils/hooks/__tests__/useStorage.test.ts
```diff
@@ -84,4 +85,72 @@ describe('useStorage', () => {
...
+    expect(getCurrentValue()).toBe(true);
+
+    // Manually mangle local storage
+    await act(async () => {
+      await setValue('ooga-booga' as unknown as boolean);
+    });
+
+    // We have no schema, so it should just return the raw JSON parsed value
+    // Not sure why anyone would want this...
+    expect(getCurrentValue()).toBe('ooga-booga');
```
evad1n, line 154: I am open to suggestion here. I can't imagine a good use case where you would want to parse JSON but not use a schema, as it would open it up to incorrect type casting.

If I were to remove that case, it seems like it would just make it simpler that if a schema is present, it will always use JSON to set and get.

And with that being said, it might be even better to enforce a schema to mitigate any possible bad data that could get put in storage

Thienhuynh95: I agree, when json is set, it should require schema. I also think it should always require schema since there is not a case where we don't need type safe, especially for local and session storage keys 🤔. But requiring it now will also add more maintenance debt. In my opinion, Amino has `ts-toolbelt` as a dev dependencies, we can utilize the `Union.Strict` to make `schema` required when `json` is true for now, and we will convert it all as we go 🤔. What do you think?

Thienhuynh95 (approved):
Looking great

src/utils/hooks/__tests__/useStorage.test.ts
evad1n: The amino hook isn't actually used that much yet so I think it is fine to require schema now.

goodpickle (approved):

Thienhuynh95 (approved):

PR #569, opened 2023-09-20
fix(filter): select enter key handler and date filters

evad1n:
## Description

This PR fixes the select filter to handle enter keypress better.

Also fixes some date filter shenanigans.

## Todo


ckgardner (dismissed):
This appears to be working great. Thanks William!

Thienhuynh95 (dismissed):

ckgardner (approved):
Now it appears to be working great!

PR #589, opened 2023-10-04
fix(types): removed ITypes from Amino

ckgardner:
## Description

I noticed in zonos.com that we were importing some types from Amino that were `ITypes`. This PR removes the I prefix from a few types.

## Todo


src/types/Option.ts
```diff
@@ -1,6 +1,6 @@
 import type { ReactNode } from 'react';
 
-export interface IOption<V extends string | number = string | number> {
+export interface Option<V extends string | number = string | number> {
   icon?: ReactNode;
```
evad1n, line 4: Can we change this to be `SelectOption` to be more specific. It is a pretty widely used type and `Option` isn't very specific.

src/components/select/CountryPhoneSelect.tsx
```diff
@@ -16,8 +16,8 @@ import { InputValuePrefix } from 'src/components/input/input-type/_FloatLabelInp
 import { Select } from 'src/components/select/Select';
 import { ChevronDownIcon } from 'src/icons/ChevronDownIcon';
 import { theme } from 'src/styles/constants/theme';
-import type { ICountryOption } from 'src/types/ICountry';
-import type { IOption } from 'src/types/IOption';
+import type { CountryOption } from 'src/types/Country';
+import type { Option as BaseOption } from 'src/types/Option';
```
evad1n, line 20: Then this can just be `SelectOption` as well. very nice

goodpickle (approved):

evad1n (approved):

evad1n (approved):

PR #602, opened 2023-10-30
Updated primary buttons disabled state

connorvang:
## Description

Updates to the primary buttons disabled states to be more obivous. 

Added a "disabled button shadow" for the new disabled states. 

## Todo

- [ ] Bump version


evad1n:
It looks like the outline button disabled state is still the same color. Is that desired?

connorvang:
Yeah thats expected. This was to only change the filled buttons. We don't have any areas where we use a disabled outline buttons that I can think of.

evad1n (dismissed):

evad1n (approved):

PR #606, opened 2023-11-13
Fe 710 research on the css module

Thienhuynh95:
## Jira issue

[Research on the css module](https://myzonos.atlassian.net/browse/FE-710)

## Description

This PR:
- adds and adjusts configuration for storybook and compiled script to use `scss modules`
- adds script to auto generated typed for scss file, when you run command `pnpm dev`, it will build the `*.d.ts` for your module `scss` and then watch changes and act accordingly with your scss changes. So we can have types-safe for importing class module from the `*.module.scss` in your component.
- converted styled component into `scss` modules for the `Layout.tsx` component.
- update docs

## Todo


src/components/layout/Layout.module.scss
```diff
@@ -0,0 +1,79 @@
+@use 'theme';
+
+.aminoLayout {
+  height: 100vh;
+  overflow: hidden;
+}
+
+.contentGrid {
+  height: theme.$amino-appbar-height;
```
Thienhuynh95, line 9: `sass` module allow to scope even more with declaring module before using any variable of that module to avoid overriding variable when importing scss variables with the same name.
Ex: we use `@use` for `theme` at top, so to use variable in the `theme`, you can use this syntax `theme.$variable`.

src/components/layout/Layout.module.scss.d.ts
```diff
@@ -0,0 +1,10 @@
+export declare const aminoLayout: string;
+export declare const content: string;
+export declare const contentGrid: string;
+export declare const footer: string;
+export declare const header: string;
+export declare const searchInputWrapper: string;
+export declare const sidebarContent: string;
+export declare const styledLogoSidebar: string;
+export declare const styledSearchInput: string;
+export declare const styledSidebar: string;
```
evad1n, line 10: Is this a manual process to create this file?

package.json
```diff
@@ -30,7 +31,7 @@
     "postbuild:dts": "echo 'done build dts'",
     "postbuild:prep:publish": "echo 'done building for publishing'",
     "postbuild:theme": "echo 'done generating styles'",
-    "postbuild": "cp -r ./src/styles/*.css ./dist && pnpm build:prep:publish",
+    "postbuild": "cp -r ./src/styles/*.css ./dist && cp -r ./src/styles/*.scss ./dist && pnpm build:prep:publish",
```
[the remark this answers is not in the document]
evad1n: yeh they use a lower level shell in gh actions

src/components/layout/Layout.module.scss.d.ts
Thienhuynh95: No, I includes the command to generate types for scss file type in the `pnpm dev` and `pnpm build`. Those two commands are already part of the developing process, so the process would not change at all, it would just magically work.

evad1n (dismissed):

evad1n (approved):

PR #610, opened 2023-11-13
feat(ProgressBar): MVP for `ProgressBar` component

nick-lanners:
## Jira issue
[FE-54 - Screen - Main page](https://myzonos.atlassian.net/browse/FE-54)

## Description

This PR implements the `ProgressBar` component that can be set to change colors as progress fills props:
- `colorStyle: greenToRed | redToGreen | Color` - controls the color of the filling bar.
- `progress: number` - percentage of bar to be filled. Component will bind the number 0-100.
- `showText: boolean` - controls whether the text representation of the progress is shown. Example: `80/100`

originally built to be used on the Screen UI page in Dashboard

## Todo


src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }};
+  z-index: 1;
+  border-radius: ${theme.radius8};
+`;
+
+export type ProgressBarProps = {
+  colorStyle?: 'greenToRed' | 'redToGreen' | Color;
+  progress?: number;
+  showText?: boolean;
+};
```
evad1n, line 59: Definitely want a `className` prop here (from `BaseProps`).

src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+          return theme.green600;
+        }
+        return theme.orange600;
+      default:
+        return getAminoColor(p.colorStyle);
+    }
+  }};
+  z-index: 1;
+  border-radius: ${theme.radius8};
+`;
```
evad1n, line 53: If future use cases come up this could possibly be refactored have color breakpoints passed as props. I think this is good for now 👍

src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+  };
+
+  const validatedProgress = validateProgress();
+
+  return (
+    <Wrapper>
+      <Base>
+        <Bar colorStyle={colorStyle} progress={validatedProgress} />
+      </Base>
+      {showText && <Text type="caption">{validatedProgress}/100</Text>}
```
evad1n, line 83: Just from looking at the design and from a practical standpoint, I think this component could just be the progress bar, and the Text can be added upstream. The design has different `Text` types in different places, and I think this component doesn't need to know about the text part.

evad1n (changes requested):
See my comments about the `className` and `Text`. Looks good otherwise 👍

Thienhuynh95 (approved):
Looks great!

evad1n (dismissed):

goodpickle (approved):

PR #613, opened 2023-11-24
refactor(Card): css modules

nick-lanners:
## Jira issue

[FE-734 - Convert `Card` to Css modules in Amino](https://myzonos.atlassian.net/browse/FE-734)

## Description

This PR converts the `Card` component to use CSS modules instead of Styled Components

## Todo

### DIFF
--- a/src/components/card/Card.tsx
+++ b/src/components/card/Card.tsx
@@ -1,59 +1,13 @@
-import type { ReactNode } from 'react';
+// Card.tsx
 
-import styled from 'styled-components';
+import type { ReactNode } from 'react';
 
 import { HStack } from 'src/components/stack/HStack';
 import { Text } from 'src/components/text/Text';
 import { theme } from 'src/styles/constants/theme';
 import type { BaseProps } from 'src/types/BaseProps';
 
-const StyledCard = styled.div<{ $spacing?: string }>`
-  border: ${theme.border};
-  border-radius: ${theme.radius6};
-  padding: ${p => p.$spacing || theme.space24};
-  background: ${theme.gray0};
-
-  /** Dividers should stretch the whole width by default */
-  hr {
-    margin-left: ${p =>
-      p.$spacing ? `calc(${p.$spacing} * -1)` : theme.spaceNegative24};
-    margin-right: ${p =>
-      p.$spacing ? `calc(${p.$spacing} * -1)` : theme.spaceNegative24};
-  }
-`;
-
-const CardHeader = styled.div<{ $spacing?: string }>`
-  margin: ${p =>
-    p.$spacing ? `calc(${p.$spacing} * -1)` : theme.spaceNegative24};
-  padding: ${p => p.$spacing || theme.space24};
-  display: flex;
-  align-items: center;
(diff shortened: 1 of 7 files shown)


src/components/card/Card.tsx
```diff
@@ -75,20 +29,38 @@ export const Card = ({
   label,
   spacing = theme.space24,
 }: CardProps) => (
-  <StyledCard $spacing={spacing} className={className || ''}>
+  <div
+    className={`${styles.card} ${className || ''}`}
```
evad1n, line 33: I think it's best to use the clsx library (i think that is what we reached consensus on using) for doing anything beyond a simple className prop. Just makes it cleaner and more consistent, and funnels className logic through one place so it is easy to refactor.

src/components/card/Card.tsx
```diff
@@ -75,20 +29,38 @@ export const Card = ({
...
+    style={{ padding: spacing }}
+  >
     {label && (
-      <CardHeader $spacing={spacing}>
+      <div
+        className={styles.cardHeader}
+        style={{
+          margin: spacing ? `calc(${spacing} * -1)` : theme.spaceNegative24,
+          marginBottom: spacing || theme.space24,
+          padding: spacing || theme.space24,
```
Thienhuynh95, line 42: We discussed about this, and we might need to refactor this to use css variable instead.

src/components/layout/Layout.tsx
```diff
@@ -31,16 +31,16 @@ export const Layout = ({
   searchInput,
   sidebar,
 }: LayoutProps) => (
-  <main className={[className, layoutStyles.aminoLayout].join(' ')}>
+  <main
+    className={[className, layoutStyles.aminoLayout].join(' ')}
+    style={{
+      '--height': headerContent ? `calc(100vh - ${theme.appbarHeight})` : '',
```
Thienhuynh95, line 37: `100vh` is better be the fallback here instead of empty string, because empty string is a valid css variable, and it will not fallback to `100vh` as you have in `var(--height, 100vh)`, the compiled css would just be *~~height: ''~~*. And then in the case you just don't want to set the value (like you want to keep , use `initial` instead of empty string to just use the default property value from the browser or `inherit` to use the style from the parent.

src/components/layout/Layout.module.scss
```diff
@@ -1,12 +1,14 @@
 @use 'theme';
 
+$height: var(--height, '100vh');
```
Thienhuynh95, line 3: The browser would only use the fallback variable just in case the variable is not defined/exist, it will not use fallback if the variable is invalid. So the default better uses in the `*.tsx` file.
```suggestion
$height: var(--height);
```

src/components/layout/Layout.tsx
nick-lanners: Ah I see. Getting my JS and CSS mixed up

Thienhuynh95 (approved):

goodpickle (dismissed):

Thienhuynh95 (approved):

PR #615, opened 2023-11-30
Refactor/css modules

nick-lanners:
## Jira issue
[FE-739 - Convert Styled Components to Css Modules: A-B](https://myzonos.atlassian.net/browse/FE-739)

## Description
This PR converts the following components to CSS Modules:
- AlertDialog
- AnnouncementDialog
- Avatar
- Badge
- Banner
- Button
  - MenuButton
  - Ripple
  - DropdownAnimation

## Testing
Domain: [css-modules.amino.zonos.com](https://css-modules.amino.zonos.com/)

1. Go to each of the components' stories listed above
2. Ensure that the styles and functionality work correctly
3. Change the variables to ensure that they are all working correctly

NOTE: Please double check the types in the `Button` component on lines 60-80 and the component return at line 306. These needed to be altered to allow for the mutable `tag` prop to be used. I got it to work, and I believe it's totally type safe, but I am inexperienced in this area.

## Todo


src/components/button/RippleGroup.tsx
```diff
@@ -1,56 +1,57 @@
...
+//     position: absolute;
+//     border-radius: 50%;
+//     opacity: ${p => p.$opacity};
+//     transform: scale(4);
+//     // prettier-ignore
+//     animation: ${p => rippleAnimation(p.$opacity)}
+//       max(${p => p.$duration}ms, 600ms) ease-out;
+//     background-color: ${p => (p.$color ? theme[p.$color] : 'currentColor')};
+//   }
+// `;
```
goodpickle, line 54:
```suggestion
```

src/components/button/RippleGroup.tsx
```diff
@@ -149,7 +154,68 @@ export const RippleGroup = React.forwardRef<RippleActions, RippleGroupProps>(
...
+
+//     React.useImperativeHandle(
+//       ref,
+//       () => ({
+//         start,
+//       }),
+//       [start],
+//     );
+
+//     return (
```
goodpickle, line 199:
```suggestion
```

goodpickle (changes requested):

src/components/announcement-dialog/AnnouncementDialog.tsx
```diff
@@ -72,16 +59,24 @@ export const AnnouncementDialog = forwardRef<
...
       >
-        <StyledImage $imageWidth={imageWidth}>{image}</StyledImage>
-        <Content spacing={8}>
+        <div
+          className={styles.styledImage}
+          style={{
+            '--image-margin': imageWidth ? '0 auto' : 'unset',
+            '--image-width': imageWidth ? `${imageWidth}px` : '100%',
+          }}
+        >
```
Thienhuynh95, line 68: I think we might want to namespace the css variable here. As it maybe conflict other component that's using the same css 🤔 Maybe we need to discuss more about it.

Thienhuynh95 (changes requested):

src/components/banner/Banner.module.scss
```diff
@@ -0,0 +1,103 @@
...
+
+.styledBanner {
+  border-radius: theme.$amino-radius-12;
+  padding: theme.$amino-space-16;
+  [class~='amino-button'] {
+    & :hover,
+    &:focus {
+      filter: brightness(1.1);
+    }
+  }
```
Thienhuynh95, line 15: Why do you need to select for the class that has a relatively similar `amino-button` className? Would `.amino-button` not work? 🤔 
```suggestion
 .amino-button {
    & :hover,
    &:focus {
      filter: brightness(1.1);
    }
  }
```

src/components/announcement-dialog/AnnouncementDialog.tsx
```diff
@@ -1,6 +1,5 @@
-import { type ReactNode, forwardRef } from 'react';
+import React, { type ReactNode, forwardRef } from 'react';
```
Thienhuynh95, line 1: I thought default `React` is no longer needed for React 18 version 🤔

src/components/button/Button.tsx
```diff
@@ -454,23 +57,30 @@ type ButtonBase = BaseProps & {
...
   ? HTMLAnchorElement
-  : T extends 'div'
-  ? HTMLDivElement
-  : HTMLButtonElement;
-
-export type ButtonProps<T extends GroupTag = 'button'> = ButtonBase &
-  Omit<ComponentPropsWithoutRef<T>, keyof ButtonBase | 'onClick'> & {
+  : T extends 'button'
+  ? HTMLButtonElement
+  : HTMLElement;
```
Thienhuynh95, line 72:
```suggestion
  ? HTMLButtonElement
  : HTMDivElement;
```

src/components/button/MenuButton.tsx
```diff
@@ -77,9 +54,9 @@ export const MenuButton = ({
   };
 
   return (
-    <Wrapper
+    <div
       ref={wrapperRef}
-      className={className}
+      className={clsx([styles.wrapper, className])}
```
Thienhuynh95, line 59: Does it need to be an array here 🤔 
```suggestion
      className={clsx(styles.wrapper, className)}
```

src/components/button/Button.module.scss
```diff
@@ -0,0 +1,374 @@
+@use 'theme';
+@use 'sass:meta';
+
+$buttonSize: var(--button-size);
+$buttonWidth: var(--button-width);
+$disabledOpacity: var(--disabled-opacity);
+$buttonPadding: var(--button-padding);
+$buttonRadius: var(--button-radius);
+$buttonFontWeight: var(--button-font-weight);
+$aminoColor: var(--amino-color, null);
```
Thienhuynh95, line 10: 🤔 I don't think color `null` is valid, so either removing it from the fallback or replace it with something else that's valid.

src/components/button/Button.module.scss
```diff
@@ -0,0 +1,374 @@
+@use 'theme';
+@use 'sass:meta';
+
```
Thienhuynh95, line 3: I see you use the `sass:meta`, and not seeing you use any `meta` mixin function.  Maybe it's not necessary here 🤔

src/components/button/Button.module.scss
```diff
@@ -0,0 +1,374 @@
...
+$buttonSize: var(--button-size);
+$buttonWidth: var(--button-width);
+$disabledOpacity: var(--disabled-opacity);
+$buttonPadding: var(--button-padding);
+$buttonRadius: var(--button-radius);
+$buttonFontWeight: var(--button-font-weight);
+$aminoColor: var(--amino-color, null);
+$backgroundColor: var(--amino-background-color, null);
```
Thienhuynh95, line 11: same as above ☝️

src/components/banner/Banner.tsx
```diff
@@ -169,106 +127,50 @@ export const Banner = ({
...
-              />
-            ),
-            removeIconColor: 'gray800',
-          })}
-        </DefaultBanner>
-      );
-  }
+  const bannerClass = intent ? styles[`${intent}Banner`] : styles.defaultBanner;
+  return (
+    <div className={clsx([styles.styledBanner, bannerClass, className])}>
```
Thienhuynh95, line 172: Does it need to be array 🤔 
```suggestion
    <div className={clsx(styles.styledBanner, bannerClass, className)}>
```

src/components/button/Button.tsx
```diff
@@ -454,23 +57,30 @@ type ButtonBase = BaseProps & {
...
+// source: https://stackoverflow.com/questions/55969769/typing-a-dynamic-tag-in-react-with-typescript#:~:text=I%20don%27t%20see,the%20div%20tag.
+export type ButtonProps<T extends GroupTag = typeof DEFAULT_TAG> =
+  ButtonBase & {
+    href?: T extends 'a' ? string : never;
     onClick?: MouseEventHandler<MyHtmlElement<T>>;
-    tag?: T;
-  };
+    tag?: T | GroupTag;
+  } & (Omit<ComponentPropsWithoutRef<T>, keyof ButtonBase | 'onClick'> &
+      HTMLAttributes<HTMLElement>);
```
Thienhuynh95, line 81: ComponentPropsWithoutRef<T> would already have all of the attributes for the `tag`. Would extending `HTMLAttributes<HTMLElement>` be necessary here?
```suggestion
  } & (Omit<ComponentPropsWithoutRef<T>, keyof ButtonBase | 'onClick'>;
```

src/components/announcement-dialog/AnnouncementDialog.tsx
nick-lanners: I don't know. Maybe it's not. Either way, with putting the `style` in `BaseProps` we don't need the `React.CSSProperties` here, so it is removed.

src/components/banner/Banner.module.scss
nick-lanners: unfortunately, that doesn't work. If Sass/css-modules sees a normal class in the module like that, it converts it to random string (ex. OL8_98nAGpDA6aY2B6A). It does this differently in each module allowing them to be scoped. So this random string will be different than the class that is on the button we are trying to style.
I've done some looking around, and it doesn't look like there is a great way to target classes from other modules like this. That's the whole point of the css-module philosophy, I guess. There are suggestions that we could use a `:global` class, but it's not ideal.

nick-lanners: EDIT: Thien and I talked and this and decided that the best way to access the `amino-button` class on the Button is to use `:global(.amino-button) {}`. This prevents post-css from transforming the class name, leaving it as normal css. We will use this syntax any time we need to access a class from another module.

src/components/button/Button.tsx
nick-lanners: It appears to be necessary. When removing the `HTMLAttributes<HTMLElement>` type, the `Tag` "component" throws an error that the types of property `onCopy` are incompatible for the anchor and button elements. When solving for that, we get another similar error for a similar property. I went down a trail of solving for about 12 of these properties before I gave up. The `HTMLAttributes<HTMLElement>` resolves that.

But this is definitely a part I could be misunderstanding. I am open to being taught about it.

src/components/button/Button.tsx
```diff
@@ -535,112 +163,167 @@ export function Button<T extends GroupTag = 'button'>({
...
+        '--amino-button-background-color': getBackgroundColor(),
+        '--amino-button-color': getColor(),
+        '--amino-button-disabled-opacity': outline ? 0.6 : 1,
+        '--amino-button-font-weight': getFontWeight(),
+        '--amino-button-hover-background-color': getHoverBackground(),
+        '--amino-button-padding': getPadding(),
+        '--amino-button-radius': getRadius(),
+        '--amino-button-size': `var(--amino-size-${size})`,
+        '--amino-button-width': props.fitContentWidth ? 'fit-content' : '',
+        ...style,
```
goodpickle, line 319: Move to the top

goodpickle (approved):

src/components/avatar/AvatarBase.tsx
```diff
@@ -43,56 +44,10 @@ export type FullAvatarProps = BaseProps & {
 
 type AvatarBaseProps = BaseProps & {
   children?: ReactNode;
+  style?: React.CSSProperties;
```
Thienhuynh95, line 47: This is already part of the BaseProps, it's not necessary since you are not overwriting it.
```suggestion
```

src/components/avatar/AvatarBase.tsx
```diff
@@ -103,21 +58,22 @@ export const AvatarBase = ({
...
+    style={{
+      ...style,
+      '--background-color': backgroundColor || `${theme.gray100}`,
+      '--background-image': `url(${backgroundUrl})`,
+      '--background-position': backgroundPosition || 'center',
+      '--background-size': backgroundSize || '',
+      '--border': bordered ? `${size / 16}px solid ${theme.gray0}` : '',
+      '--border-radius': avatarShapes[shape],
+      '--height': `${size}px`,
+      '--width': `${size}px`,
```
Thienhuynh95, line 74: Should this be namespaced like `--amino-avatar-base-background-color`?

Thienhuynh95 (commented):
Looks good, just some more suggestions 😉

src/components/avatar/AvatarBase.tsx
nick-lanners: Ah! I thought I got them all. good catch

src/components/avatar/AvatarBase.tsx
nick-lanners: Yes it should. thanks

goodpickle (dismissed):

Thienhuynh95 (dismissed):

Thienhuynh95 (approved):

goodpickle (approved):

PR #624, opened 2023-12-11
refactor(css): convert to css modules

nick-lanners:
## Jira issue
[FE-750 - Convert to CSS Modules - L-P](https://myzonos.atlassian.net/browse/FE-750)

## Description
This PR convert the following to css modules:
- NavigationGroup
- CollapsibleList
- List
- ListItem
- Menu
- MenuItem
- _Filter
- _TableData
- NestedDataTable
- PivotTable
- ProfileIcon
- ProgressBar

## Todo


src/components/list-item/ListItem.tsx
```diff
@@ -118,26 +50,28 @@ export const ListItem = forwardRef<HTMLDivElement, Props>(
     },
     ref,
   ) => (
-    <AminoListItem
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+    <div
       ref={ref}
-      className={className}
-      disabled={disabled}
+      className={clsx(
```
Thienhuynh95, line 56: I see quite a few places having this comment eslint disable `jsx-a11y/no-static-element-interactions`. To not to have this comment, as the rule suggest [here](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md#succeed), you may want to add `role="button"` to any static element that has `onClick` event on it.
Ex: 
```
<div
    role="button"
    ref={ref}
    ...
```

src/components/menu/MenuItem.tsx
```diff
@@ -46,9 +17,15 @@ export const MenuItem = ({
...
   onClick,
+  style,
 }: MenuItemProps) => (
-  <StyledListItem className={className} onClick={onClick}>
+  // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
+  <li
+    className={clsx(className, styles.styledListItem)}
+    onClick={onClick}
+    style={style}
+  >
```
Thienhuynh95, line 27: You may remove the eslint disable here too if adding`role="button"` to the element
```suggestion
  // eslint-disable-next-line jsx-a11y/click-events-have-key-events
  <li
    role="button"
    className={clsx(className, styles.styledListItem)}
    onClick={onClick}
    style={style}
  >
```

src/components/menu/MenuItem.tsx
```diff
@@ -46,9 +17,15 @@ export const MenuItem = ({
   className,
   icon,
   onClick,
+  style,
 }: MenuItemProps) => (
-  <StyledListItem className={className} onClick={onClick}>
+  // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
```
Thienhuynh95, line 22: We may want to disable the `jsx-a11y/click-events-have-key-events` rules since we don't always have the key events for each click event now. We might want to in the future, but I don't think now is the time, especially since we are migrating to `css-modules`.

src/components/list-item/ListItem.tsx
nick-lanners: When doing this, I still get "elements with click handlers must have at least one keyboard listener. Should I add `onKeyDown={() => {}}` or listen for the enter key to 'click' or something else?

goodpickle (approved):

Thienhuynh95 (approved):

PR #634, opened 2023-12-28
fix(Filter): ref warning

evad1n:
## Description

This PR fixes the filter ref error.

## Todo

### DIFF
--- a/src/components/filter/FilterWrapper.tsx
+++ b/src/components/filter/FilterWrapper.tsx
@@ -23,7 +23,6 @@ type FilterWrapperProps = BaseProps & {
   filterText: string;
   hasFilter: boolean;
   label: string;
-  ref: RefObject<HTMLDivElement>;
   handleApply: () => void;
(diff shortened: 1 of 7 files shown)


nick-lanners (dismissed):

src/components/filter/FilterWrapper.tsx
```diff
@@ -44,11 +43,9 @@ export const FilterWrapper = ({
   handleToggle,
   hasFilter,
   label,
-  ref,
   style,
```
Thienhuynh95, line 46: You can use React.forwardRef to use ref unless we don't have anywhere pass the ref down to this component and you want to take the ref out completely.

evad1n: Yeh it wasn't being used at all apparently

Thienhuynh95 (approved):

ckgardner (approved):

PR #646, opened 2024-01-15
fix: checkbox className

nick-lanners:
## Jira issue
Related to [FE-810 - Dashboard to css modules](https://myzonos.atlassian.net/browse/FE-810)

## Description
This PR adds a `'disabled'` string to a disabled checkbox classList becuase tests in Dashboard depend on it

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.1.89",
+  "version": "5.1.90",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:Zonos/amino.git",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


goodpickle (dismissed):

goodpickle (approved):

evad1n:
Weird that the test didnt catch this...

evad1n:
Ah, the amino test is a regex test so it would match, I'm guessing the `classList.contains` upstream didn't satisfy that. Maybe the amino regex should be more specific...

nick-lanners:
Yeah the Dashboard test is in `FedExGroundCollect.test.tsx` and it's `classList.contains('disabled') which I imagine has to be literally 'disabled' with no other nonsense from the css module. This change fixed it.

PR #650, opened 2024-01-18
fix: undo class passing to input

nick-lanners:
## Description
This PR fixes a mistake from the previous version.
I misunderstood what I was looking at in the Input component. The previous "fix" ends up passing the classname twice, which explodes in the input when it is styled.

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.1.94",
+  "version": "5.1.95",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:Zonos/amino.git",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


evad1n:
Maybe this is the correct way, is our standard to pass the `className` to the outermost tag? In that case the other `className` in the `renderInput` should be removed maybe.

nick-lanners:
@evad1n It is tricky because the component was originally built with the className passed through the `renderInput`, so moving them would mess up a bunch of styled input components.

The `style` prop should go through the outermost tag, but it seems this one had the class go through the renderInput...?

evad1n (dismissed):

goodpickle (approved):

PR #659, opened 2024-02-06
adjusted colors for visual brightness and some other style tweaks

connorvang:
## Description

Updated colors for more accurate visual brightness with the tints and shades

Added a "raised-surface" variable for buttons and inputs so that in dark mode they appear lighter on raised elements like dialogs.

Tweaked the small header type to allow for sentence case and not always uppercase

## Todo


build-utils/css/constants/theme/day/colors/_blue.ts
```diff
@@ -1,22 +1,22 @@
 export const blue = {
   /** @info #DEEDFE  */
-  'blue-100': '#DEEDFE',
+  'blue-100': '#E7F6FE',
```
Thienhuynh95, line 3: Can you also update the comment jsdocs above to match with the value of the variable?

Thienhuynh95 (changes requested):

build-utils/css/constants/theme/day/colors/_blue.ts
connorvang: Ah yes. Ill do that

evad1n (approved):

Thienhuynh95 (approved):

PR #665, opened 2024-02-12
fixed the dialog on the menu button and adjusted other borders

connorvang:
## Description

Fixed the menu for the menu button being the wrong color (my bad) and adjusted other borders I I missed earlier.

## Todo

### DIFF
--- a/src/components/simple-table/SimpleTable.module.scss
+++ b/src/components/simple-table/SimpleTable.module.scss
@@ -4,10 +4,10 @@
   width: 100%;
 
   > thead {
-    text-transform: uppercase;
+    text-transform: none;
 
     > tr {
-      border-bottom: theme.$amino-border;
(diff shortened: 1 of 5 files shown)


src/components/simple-table/SimpleTable.module.scss
```diff
@@ -4,10 +4,10 @@
   width: 100%;
 
   > thead {
-    text-transform: uppercase;
+    text-transform: none;
```
evad1n, line 7: What's the reason for this change? In all the designs I have seen the headers are always capitalized.

connorvang: We are slowly moving away from all caps across the board and this was one of the last remaining places in Amino where that text-transform is applied.

evad1n (approved):

evad1n (approved):

PR #670, opened 2024-02-14
feat(coversheet): make close button optional

goodpickle:
## Linear issue

[Implement iframe for Screen calculator UI](https://linear.app/zonos/issue/TS-530/implement-iframe-for-screen-calculator-ui)

## Description

This PR makes the close button optional. This will allow us to use Coversheets without a close button in iframes.

## Todo

### DIFF
--- a/src/components/cover-sheet/CoverSheet.tsx
+++ b/src/components/cover-sheet/CoverSheet.tsx
@@ -20,6 +20,7 @@ export type CoverSheetProps = BaseDialogProps & {
   actions?: ReactNode;
   headerComponent?: ReactNode;
   label: string;
+  showCloseButton?: boolean;
 };
 
 export const CoverSheet = ({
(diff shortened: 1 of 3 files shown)


src/components/cover-sheet/CoverSheet.tsx
```diff
@@ -31,6 +32,7 @@ export const CoverSheet = ({
   label,
   onClose,
   open,
+  showCloseButton = true,
```
evad1n, line 35: What do you think about making this boolean prop default to `false`? I probably didn't follow this rule in the past, but I have come to see that it is more idiomatic to have boolean props default to false, so when you want to change them you can use the shorthand of just the prop name.

```
<
  hideCloseButton
/>
```

```
< 
  showCloseButton={false}
/>
```

evad1n (dismissed):

evad1n (approved):

PR #674, opened 2024-03-01
Fix/useStorage-avoid-dependency-array-issues

goodpickle:
## Description

This PR updates the useStorage hook to use a callback for the setValue, so it doesn't trigger a change in the dependency array of the consuming effects.

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.1.116",
+  "version": "5.1.117",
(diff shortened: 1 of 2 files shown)


evad1n:
Looks like we jumped 2 versions 🤷

evad1n (dismissed):

goodpickle (commented):
Yeah, I tried bumping the version before pushing. I just added a fix for that.

evad1n (approved):

PR #682, opened 2024-03-20
Build/upgrade storybook

evad1n:
## Linear issue

completes https://linear.app/zonos/issue/FE-792/upgrade-amino-to-use-storybook-8

## Description

This PR upgrades storybook to the latest (v8) and switches from webpack to vite.

Previously we kept using webpack, despite vite being faster, because we wanted to mimic the same setup as NextJs, which used webpack internally. They are moving on to using turbopack now, so that logic is no longer valid.

Other changes:
- change the storybook dev script in package.json to be `storybook:dev`, as otherwise it is impossible to run the `storybook` script in the pnpm environment with args (it would just run our `storybook` script).
- removes unnecessary packages (webpack, extra sass stuff).
- Replaces some mistaken storybook imports (`useState` from storybook instead of react)

## Todo


vite.config.ts
Thienhuynh95, line 1: Intestesting, vitest can use the same vite config file? That's convenient!

Thienhuynh95 (dismissed):

vite.config.ts
evad1n: Yeah, vitest config is actually just a vite config.

Thienhuynh95 (approved):

Thienhuynh95 (approved):

PR #687, opened 2024-04-11
Fix/more ssr errors

evad1n:
## Description

This PR fixes some more SSR errors encountered within the app router.

- `Toast`: key spreading (can't spread the `key` prop in react)
- `ThemeSelect`: server was always day (default value), which caused some class conflicts in stuff like the toggle, so added logic to only run on client

also: 
- password input eye icon has a visible focus state now (was bad UX when using keyboard navigation and you can't tell what element is focused)
- remove `inputPrefix` and `inputSuffix` from `Input`. This was logic related to upstream usage of `NumberInput`, which shouldn't be in amino

## Todo

### DIFF
--- a/src/components/input/Input.tsx
+++ b/src/components/input/Input.tsx
@@ -134,7 +134,7 @@ export const Input = ({
             readOnly={readOnly}
             required={required}
             size={size}
-            suffix={suffix === null ? null : suffix}
+            suffix={suffix}
             tabIndex={tabIndex}
             value={value || ''}
             valuePrefix={valuePrefix}
--- a/src/components/input/Input.tsx
+++ b/src/components/input/Input.tsx
@@ -2,7 +2,6 @@ import type {
   ChangeEventHandler,
   InputHTMLAttributes,
   MutableRefObject,
-  ReactNode,
 } from 'react';
 
 import clsx from 'clsx';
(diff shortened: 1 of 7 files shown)


src/components/input/Input.tsx
```diff
@@ -46,9 +43,7 @@ export const Input = ({
   error,
   helpText,
   inputMode,
-  inputPrefix,
```
goodpickle, line 49: This was a necessary work-around for the `<NumberInput />` in Dashboard. It looks like we are only using the `<NumberInput />` in a few places that are older. One option is to kill the `<NumberInput />` in Dashboard, but that may be a rabit hole.

```
/**
 * @desc An input that takes a number & returns a number
 * @tutorial https://github.com/s-yadav/react-number-format
 * @param `prefix` uses `react-number-format` prefix
 * @param `suffix` uses `react-number-format` suffix
 * @param `inputPrefix` uses `@zonos/amino` prefix
 * @param `inputSuffix` uses `@zonos/amino` suffix
 */
export const NumberInput = <T extends boolean = false>({
```

goodpickle (changes requested):

src/components/input/Input.tsx
evad1n: I understand the repercussions in dashboard and will have a separate PR to fix the resulting issue there. I don't think that logic belongs in amino.

goodpickle: May I ask what your plan is? `suffix` and `prefix` props are absorbed by the `react-number-format` package, which would deem a different prop necessary.

evad1n: My idea is to just move this logic into dashboard with a wrapper component.

goodpickle (dismissed):

goodpickle (approved):

PR #689, opened 2024-04-16
fix(useCountryOptions): Updated useCountryOptions to work for all projects

ckgardner:
## Description

This PR updates the `useCountryOptions` hook so it can be used in our other projects. It uses localStorage as the first source of data. If it doesn't exist, fetch using SWR. It will then save it to localStorage and SWR cache. Note: It applies the same `no cache` header that [Will applied to zonos.com](https://github.com/Zonos/zonos.com/pull/1588) yesterday.

This hook now takes a single parameter of `dashboardUrl` in order to reference our API correctly from any environment in any project. For the stories in amino, I extracted the `getCountriesUrl` into a util for storybook purposes. It also takes a generic `TCountryCode` type that we will need to pass in from our various projects since this hook has no reference our graph CountryCode type.

## Preview

<https://countries.amino.zonos.com/?path=/story/amino-select-countrymultiselect--basic-country-multi-select>

Visit any of the country selects in the url above

## Todo


src/utils/storage.ts
```diff
@@ -61,10 +64,36 @@ export const setStorageItem = async <Value extends unknown>({
...
-  storage.setItem(key, valueToSet);
+  if (isClientSide && typeof valueToSet === 'string') {
+    storage.setItem(key, valueToSet);
+    storage.setItem(`${key}_update_after`, getThisWeek());
+    await mutate(key, valueToSet, false);
+  } else if (isClientSide) {
+    storage.removeItem(key);
+    storage.removeItem(`${key}_update_after`);
+    await mutate(key, null, false);
+  }
```
evad1n, line 75: I don't think this code makes sense in this context anymore. This code specifically will always satisfy the first condition (it gets stringified if it is not a string). The else if condition will never run. 

I think this should be removed out of here and be done in the hook instead. Also I realize that we aren't actually using SWR for this amino hook now anyway so the mutate can probably be removed.

evad1n (changes requested):
.

src/utils/storage.ts
```diff
@@ -65,6 +70,51 @@ export const setStorageItem = async <Value extends unknown>({
...
+  lifetime,
+  type,
+  value,
+}: SetParams<Value> & { lifetime: Dayjs }) => {
+  const storage =
+    type === 'session' ? window.sessionStorage : window.localStorage;
+
+  const valueToSet = typeof value === 'string' ? value : JSON.stringify(value);
+
+  const valueIsTruthy = !!value;
```
evad1n, line 86: My only concern is maybe making this function a parameter, as this truthy logic may not always be desired.

ckgardner (commented):
This looks good. Have you tested this solution?

src/components/filter/__stories__/Filter.stories.tsx
```diff
@@ -16,11 +17,11 @@ import {
 import { FilterMultiSelect } from 'src/components/filter/filter-multi-select/FilterMultiSelect';
 import { FilterSelect } from 'src/components/filter/filter-select/FilterSelect';
 import { FilterText } from 'src/components/filter/filter-text/FilterText';
-import { useCountryOptions } from 'src/components/select/__stories__/useCountryOptions';
 import { type Flag, FlagIcon } from 'src/icons/flag-icon/FlagIcon';
 import { RemoveIcon } from 'src/icons/RemoveIcon';
-import type { CountryOption } from 'src/types/Country';
 import type { SelectOption } from 'src/types/SelectOption';
+import type { CountryOption } from 'src/utils/hooks/useCountryOptions';
+import { useCountryOptions } from 'src/utils/hooks/useCountryOptions';
```
Thienhuynh95, line 24: I'm surprised it didn't merge the type into one 🤔 Is this the new eslint rule? Disregard if it's actually is 😄 
```suggestion
import { type CountryOption, useCountryOptions } from 'src/utils/hooks/useCountryOptions';
```

src/components/select/__stories__/CountryMultiSelect.stories.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { getCountryUrls } from 'story-utils/getCountryUrls';
 
-import { useCountryOptions } from 'src/components/select/__stories__/useCountryOptions';
 import {
   type CountryMultiSelectProps,
   CountryMultiSelect,
 } from 'src/components/select/CountryMultiSelect';
-import type { CountryOption } from 'src/types/Country';
+import type { CountryOption } from 'src/utils/hooks/useCountryOptions';
+import { useCountryOptions } from 'src/utils/hooks/useCountryOptions';
```
Thienhuynh95, line 11: Same here 🤔

src/components/select/__stories__/CountrySelect.stories.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { getCountryUrls } from 'story-utils/getCountryUrls';
 
-import { useCountryOptions } from 'src/components/select/__stories__/useCountryOptions';
 import {
   type CountrySelectProps,
   CountrySelect,
 } from 'src/components/select/CountrySelect';
-import type { CountryOption } from 'src/types/Country';
+import type { CountryOption } from 'src/utils/hooks/useCountryOptions';
+import { useCountryOptions } from 'src/utils/hooks/useCountryOptions';
```
Thienhuynh95, line 11: Same here

.storybook/utils/getCountryUrls.ts
```diff
@@ -0,0 +1,8 @@
+export const getCountryUrls = () => {
+  const dashboardUrl = import.meta.env.STORYBOOK_ZONOS_DASHBOARD_URL || null;
```
Thienhuynh95, line 2: storybook use `import.meta.env` now !? 🤔 Interesting, why wouldn't we use `environment.mjs`?

ckgardner: We had this `import.meta` for a while now. We just haven't done `environment.mjs` yet because of how few variables we have.

evad1n: with the vite change. That is the way vite uses imports. We only have one env variable in amino (we have another for scripts) so no need for something heavy like that (yet)

src/components/select/__stories__/CountrySelect.stories.tsx
evad1n: Hmmm interesting, we may not have that rule in amino.

evad1n: I will look into this

.storybook/utils/getCountryUrls.ts
evad1n: If we were to use it here, we just change the runtime values to use the `import.meta.env`

Thienhuynh95: I forgot this is Amino 😅

ckgardner (commented):
Great PR

Thienhuynh95 (approved):

evad1n:
Just testing upstream in dashboard to make sure it works.

src/components/connection-map/ConnectionMap.tsx
```diff
@@ -14,9 +14,9 @@ import { feature } from 'topojson-client';
 import { Skeleton } from 'src/components/skeleton/Skeleton';
 import { theme } from 'src/styles/constants/theme';
 import type { BaseProps } from 'src/types/BaseProps';
-import { type CountryOption } from 'src/types/Country';
 import { type GeoJsonWorld } from 'src/types/GeoJsonWorld';
 import { getCountryCodeByName } from 'src/utils/getCountryCodeByName';
+import { type CountryOption } from 'src/utils/hooks/useCountryOptions';
```
Thienhuynh95, line 19: I think we should turn on the no-side-effect rule for amino, looks like we only use CountryOption type, but import the whole hook in. I doubt that it would increase the size significantly 🤔

evad1n: Hmm, can we do this in a separate PR?

Thienhuynh95: yah, just point it out, that's why I didn't request changes 😆

Thienhuynh95 (approved):

evad1n (approved):

PR #690, opened 2024-04-17
Refactor/linting

evad1n:
## Description

This PR upgrades the linting rules.

See https://github.com/Zonos/amino/pull/689#discussion_r1569154862

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "eslint-plugin-prettier": "^5.0.0",
     "eslint-plugin-react": "7.32.2",
     "eslint-plugin-react-hooks": "4.6.0",
-    "eslint-plugin-simple-import-sort": "^9.0.0",
+    "eslint-plugin-simple-import-sort": "^12.1.0",
     "eslint-plugin-sort-destructure-keys": "^1.5.0",
     "eslint-plugin-sort-keys": "^2.3.5",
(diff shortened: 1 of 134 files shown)


package.json
```diff
@@ -135,7 +135,7 @@
     "eslint-plugin-prettier": "^5.0.0",
     "eslint-plugin-react": "7.32.2",
     "eslint-plugin-react-hooks": "4.6.0",
-    "eslint-plugin-simple-import-sort": "^9.0.0",
+    "eslint-plugin-simple-import-sort": "^12.1.0",
```
Thienhuynh95, line 138: The version after 9.0.0, the rule will sort for both non-types and types alphabetically and not separate sorting them like the version 9.0.0.
Ex:
9.0.0
```
import { 
  abnomal,
  best,
  type Abnormal,
  type Abtest,
  type Best
}
```
And version > 9
```
import {
  abnormal,
  type Abnormal,
  type Abtest,
  best,
  type Best,
```
Personally, I like the first one, but it's just my opinion 😅

Thienhuynh95 (dismissed):

package.json
evad1n: Hmmm, I also like the first one...

nick-lanners (approved):

Thienhuynh95 (approved):

PR #691, opened 2024-04-29
Fix: Added missing props to Flex

ckgardner:
## Description

This PR brings feature parity to the Amino Flex from the dashboard Flex component

## Todo

### DIFF
--- a/src/components/flex/Flex.tsx
+++ b/src/components/flex/Flex.tsx
@@ -53,7 +53,7 @@ export const Flex = ({
   flexWrap = 'nowrap',
   gap = 8,
   justifyContent = 'flex-start',
-  padding,
+  padding = 0,
   style,
 }: FlexProps) => (
   <div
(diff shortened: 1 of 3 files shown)


Thienhuynh95 (approved):

src/components/flex/Flex.tsx
```diff
@@ -42,7 +61,9 @@ export const Flex = ({
     style={{
       ...style,
       '--dashboard-flex-wrapper-align-items': alignItems,
+      '--dashboard-flex-wrapper-flex-children': childrenFlex,
       '--dashboard-flex-wrapper-flex-direction': flexDirection,
+      '--dashboard-flex-wrapper-flex-wrap': flexWrap,
       '--dashboard-flex-wrapper-gap': `${gap}px`,
       '--dashboard-flex-wrapper-justify-content': justifyContent,
       '--dashboard-flex-wrapper-padding': `${padding || 0}px`,
```
evad1n, line 69: What do you think about setting the default in the function parameters like the rest? (I know I was probably the one that did this in the first place)

evad1n:
I'm not seeing any modifications to the SCSS file?

src/components/flex/Flex.tsx
ckgardner: Didn't see that. Good catch

PR #693, opened 2024-05-03
feat(inlineLink): added inlineLink variant with no padding

ckgardner:
## Description

This PR adds an `inlineLink` button to allow us to do inline links that have no padding and can still use icons and other button props.

PREVIEW: <https://amino-git-feat-inline-link-zonos.vercel.app/?path=/story/amino-button--inline-link-button>

## Todo

### DIFF
--- a/src/components/button/Button.module.scss
+++ b/src/components/button/Button.module.scss
@@ -370,6 +370,22 @@ $text-button-disabled-color: var(--amino-button-text-button-disabled-color);
     }
   }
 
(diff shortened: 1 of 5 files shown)


evad1n (approved):
What do you think of changing the inline link story to demonstrate our dashboard use case instead of the other button templates?

evad1n (changes requested):
The story thing I said

evad1n (approved):

PR #694, opened 2024-05-06
Feat/filter url state

evad1n:
https://url-state.amino.zonos.com/ (for country filters)

## Linear issue

https://linear.app/zonos/issue/FE-789/preserve-order-list-filter-state-in-the-url

## Description

This PR creates a bunch of utilities for storing state in the URL.

- upgrades pnpm to v9.x

## Todo


src/utils/hooks/useStateUrl.ts
```diff
@@ -0,0 +1,106 @@
...
+    } else {
+      urlParams.set(name, valueToSet);
+    }
+  } finally {
+    if (urlParams.toString() === '') {
+      // Just remove the ? if there are no params
+      window.history.replaceState(null, '', window.location.pathname);
+    } else {
+      window.history.replaceState(null, '', `?${urlParams.toString()}`);
+    }
```
ckgardner, line 79: What happens if we have other query param logic on a page?

ckgardner (approved):

ckgardner (approved):

PR #695, opened 2024-05-06
fix(layout): auto sidebar footer height

ckgardner:
## Description

This PR removes the strict height on the amino Layout footer. Not sure why it was necessary in the first place? 🤔

I'm adding a new piece to the dashboard sidebar footer and this is what I need done.

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.2.9",
+  "version": "5.2.10",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:Zonos/amino.git",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


evad1n (approved):
I would say check upstream projects, but dashboard is probably the only place this is relevant for anyways.

ckgardner:
> I would say check upstream projects, but dashboard is probably the only place this is relevant for anyways.

Explore will be okay. Arrakis will just need the footer children height set. Not sure if any other projects use the layout?

ckgardner:
I reverted thinking a `minmax` would help for the other projects, but they both just have `<UserMenu />` components that are set to `100%` width. Those will just need to be set on a per project basis.

evad1n (approved):

PR #699, opened 2024-05-15
fix: amount filter logic

evad1n:
## Description

This PR fixes the amount filter logic.

I have decided to leave the `useStateUrl` code because it is helpful for testing the filters with state derived from outside the component, even though that code won't be used.

## Todo

### DIFF
--- a/src/components/filter/filter-amount/FilterAmount.tsx
+++ b/src/components/filter/filter-amount/FilterAmount.tsx
@@ -2,6 +2,7 @@ import {
   type Dispatch,
   type KeyboardEvent,
   useCallback,
+  useMemo,
   useState,
 } from 'react';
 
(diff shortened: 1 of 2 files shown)


src/components/filter/filter-amount/FilterAmount.tsx
```diff
@@ -38,53 +39,53 @@ export const FilterAmount = ({
...
-          setFilterText(`Equal to ${amount1String}`);
+          setFilterText(`Equal to ${amountMinString}`);
           break;
         case 'greater':
-          setFilterText(`Greater than ${amount1String}`);
+          setFilterText(`Greater than ${amountMinString}`);
           break;
         case 'less':
-          setFilterText(`Less than ${amount1String}`);
+          setFilterText(`Less than ${amountMaxString}`);
```
ckgardner, line 82: 🏅

src/components/filter/filter-amount/FilterAmount.tsx
```diff
@@ -97,16 +98,22 @@ export const FilterAmount = ({
...
+            amountMin: editingAmountMin,
+          });
           break;
         case 'greater':
-          dispatchAmounts({ amount1: editingAmount1, amount2: null });
+          dispatchAmounts({ amountMax: null, amountMin: editingAmountMin });
           break;
         case 'less':
-          dispatchAmounts({ amount1: null, amount2: editingAmount1 });
+          dispatchAmounts({ amountMax: editingAmountMax, amountMin: null });
```
ckgardner, line 116: 🏅

ckgardner (dismissed):

ckgardner (approved):

src/components/filter/filter-amount/FilterAmount.tsx
```diff
@@ -38,53 +39,53 @@ export const FilterAmount = ({
...
     },
     [dispatch],
   );
 
   const handleApplyFilterText: FilterApplyCallback = useCallback(
     setFilterText => {
-      const amount1String = `${editingAmount1?.toFixed(2) || 0}` || '';
-      const amount2String = `${editingAmount2?.toFixed(2) || 0}` || '';
+      const amountMinString = `${editingAmountMin?.toFixed(2) || 0}` || '';
+      const amountMaxString = `${editingAmountMax?.toFixed(2) || 0}` || '';
```
Thienhuynh95, line 69: I think it would never fall back to ''. Because you are trying a fallback with a string interpolation, it would always fall back to '0' if editingAmount is `null`.

PR #712, opened 2024-07-18
Polish updates

connorvang:
## Description

Adjusts the following:

- Adds "50" value to all colors to match grayscale for better contrast consistency.
- Tweaks main components with the updated color scales.
- Adjusts badges, banners, and buttons with updated colors
- Fixes to the filter component to not have an extra line when inactive
- Removes "duotone" icons from the component previews since we don't use them often and have contrast issues. switching to lines only.

## Todo


evad1n:
We have quite a few places in dashboard where we are using colors that are being removed in this PR (`gray1200`) 

image

I assume those will become `gray1000`? Just want to have clarity on how to resolve breaking changes upstream.

connorvang:
@evad1n yes it will

evad1n (approved):
Looks good. This should probably be `version:minor` instead of `version:patch`. I also would want @Thienhuynh95 's review as he is more familiar with the theme generation.

build-utils/css/constants/theme/day/colors/_blue.ts
```diff
@@ -1,22 +1,24 @@
...
-  'blue-600': '#445EEE',
-  /** @info #3843C2  */
-  'blue-700': '#3843C2',
-  /** @info #2E3D94  */
-  'blue-800': '#2E3D94',
-  /** @info #212F68  */
-  'blue-900': '#212F68',
-  /** @info #152141  */
-  'blue-1000': '#152141',
+  'blue-400': '#678FF3',
```
Thienhuynh95, line 11: Can you also update the jsdoc comment to match with the new color value?

build-utils/css/constants/theme/night/colors/_gray.ts
```diff
@@ -1,30 +1,28 @@
...
+  'gray-800': '#DFE2E7',
+  /** @info #E8EAEE  */
+  'gray-900': '#E8EAEE',
+  /** @info #F4F5F6  */
+  'gray-950': '#F4F5F6',
   /** @info #F6F6F8  */
-  'gray-1100': '#F6F6F8',
-  /** @info white  */
-  'gray-1200': '#FFFFFF',
+  'gray-1000': '#FFFFFF',
```
Thienhuynh95, line 27: Can you update the comment with the new value?

build-utils/css/constants/theme/day/colors/_blue.ts
connorvang: Ah yes. I thought I did but must have missed some

src/components/filter/FilterWrapper.module.scss
```diff
@@ -27,24 +27,28 @@ $border-right-color: var(--amino-filter-wrapper-border-right-color);
...
   align-items: center;
-  gap: theme.$amino-space-8;
-  padding: theme.$amino-space-4 theme.$amino-space-8;
+  gap: theme.$amino-space-0;
+  padding: theme.$amino-space-4 theme.$amino-space-4;
   border-radius: 100px;
   border-top-right-radius: $border-top-right-radius;
   border-bottom-right-radius: $border-bottom-right-radius;
-  border-right: 1px solid $border-right-color;
+  border-right: 0px solid $border-right-color;
```
Thienhuynh95, line 35: This seems to have no effect here. `border` with 0 width means no border. 🤔 If that's what you mean, you can just remove this line.

Thienhuynh95 (approved):

PR #714, opened 2024-07-19
FEAT: Button Group component

ckgardner:
## Description

This PR adds a button group component to Amino. It takes children of buttons

## Todo


src/components/button/__stories__/ButtonGroup.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <Button icon={<EditIcon />} />
+        <Button icon={<RemoveCircleIcon />} />
+        <Button icon={<ArrowSwapIcon />} />
+        <Button icon={<StarsIcon />} />
+      </ButtonGroup>
+    </VStack>
+  </VStack>
+);
+
+export const Default = Template.bind({});
```
evad1n, line 70: What do you think about using the new recommended storybook object form for stories:

https://github.com/Zonos/amino/blob/main/src/components/tooltip/__stories__/Tooltip.stories.tsx

https://storybook.js.org/docs/api/csf

ckgardner: Thanks for the suggestion!

evad1n (dismissed):

evad1n (approved):

PR #723, opened 2024-08-26
feat(invertedButton): Inverted button variant and fix theme override

ckgardner:
## Description

This PR adds a new button variant of `Inverted` which basically inverts the default button by changing the theme with a fixed `gray0` background color. It also fixes the `themeOverride` prop as it wasn't working properly on the story.

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.3.8",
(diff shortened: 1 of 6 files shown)


evad1n (approved):
STELLAR WORK

evad1n (approved):

PR #726, opened 2024-08-28
Fix: Final button fix for themes and inverted variant

ckgardner:
## Description

This PR officially fixes the button themes and variants. It also removes the use of the amino theme hook on the button and just uses CSS to transform the white spinners for the appropriate buttons. Screenshots of running zonos.com locally against the change.

I have learned that using the `useAminoTheme` hook within amino is something we should mostly steer away from because in instances like zonos.com we hard code the `data-theme` attribute to `night` which doesn't change what value `useAminoTheme` pulls from local storage.

Screenshot 2024-08-28 at 3 28 51 PM
Screenshot 2024-08-28 at 3 28 33 PM
Screenshot 2024-08-28 at 3 28 03 PM
Screenshot 2024-08-28 at 3 27 46 PM

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.3.11",
+  "version": "5.3.12",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:Zonos/amino.git",
   "license": "MIT",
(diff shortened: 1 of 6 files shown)


evad1n:
@evad1n you should fix the tooltips to not do this as well. `ThemeSelect` is fine still because that is how it works....

evad1n (dismissed):

evad1n (approved):

PR #730, opened 2024-09-11
feat: cleanup time everybody do your share

evad1n:
## Description

This PR:

- removes usage of `useAminoTheme` internally (modifies `Tooltip`)
- make boolean props consistently false (breaking prop name changes)
- remove `CountryPhoneSelect`
- fix a bunch of semantic HTML and a11y problems
- upgrade storybook to latest
- add sticky header to `SimpleTable`
- fix `CountryMultiSelectExpanded` country sorting (and story)
- removes `SearchInput` (this can trivially be replaced with the normal `Input` and `valuePrefix`)
- restructure `Input` to be semantically correct

Most changed components:
- Input (all of them, changing sizes, etc.)
- Checkbox
- ListItem
- Radio
- SimpleTable

Involves breaking changes upstream, which I have locally tested against dashboard and zonos.com.

Unfortunately there is no easy to make a preview domain for dashboard with these changes.

The local caddy config breaks the playwright tests if you try to run locally, so that will need to be fixed in the future. It works in CI still and caddy and vite/storybook hate each other so im procrastinating on that...

## Todo
- [x] test against dashboard
- [x] test against zonos.com
- [ ] Bump version (minor) and add tag


nick-lanners (approved):
You're doing the lord's work

nick-lanners (approved):

PR #747, opened 2024-10-16
FIX: Simple table styling and truncate text

ckgardner:
## Preview

https://amino-git-fix-components-zonos.vercel.app/?path=/story/amino-simpletable--custom

1. Click the `Disable truncate` checkbox
2. Click the `View one row` checkbox and then click the menu to see it's no longer cut off. 

## Description

This PR makes table cells truncate by default with the option to to disable truncating by column. It will just truncate at the percentage width specified in the header columns.

It also removes `overflow: auto` from the table wrapper. I didn't see any repercussions, but maybe I missed something.

## Todo

### DIFF
--- a/src/components/simple-table/SimpleTable.module.scss
+++ b/src/components/simple-table/SimpleTable.module.scss
@@ -2,7 +2,6 @@
 
 .tableContainer {
   width: 100%;
-  overflow: auto;
 }
 
 .tableStyled {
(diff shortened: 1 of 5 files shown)


goodpickle (dismissed):

src/components/simple-table/SimpleTable.module.scss
```diff
@@ -2,7 +2,6 @@
 
 .tableContainer {
   width: 100%;
-  overflow: auto;
```
evad1n, line 5: I probably added this when I was doing the sticky header. If it still works in dashboard then that's fine.

evad1n (approved):
I wonder if we could automatically add tooltips to anything that gets truncated. I believe Thien did something like that in dashboard `TruncateText`

Note to whoever is interested: if we do this then `SimpleTable` imports should probably be dynamic as it has grown quite large.

PR #748, opened 2024-10-17
FIX: Truncating simple table cell links

ckgardner:
## Linear issue

[ONB-969: Frontend - Fix minor UI bugs](https://linear.app/zonos/issue/ONB-969/frontend-fix-minor-ui-bugs)

## Description

This PR fixes simple table cells to truncate and apply other text wrapping properties. I added a `textWrapMethod` in order to specify if you want truncating, no wrap, or normal wrapping. We want our tables to default to be no-wrapping single line, so `nowrap` is the default.

This is NOT a perfect solution. Truncating will require a set percentage width in most circumstances. Only column per table can have truncating text without a `width` prop being passed because the `max-width: 0` screws it up.

I think the best solution to stop dealing with table styling is use a custom flex or grid table. I've seen other popular libraries do this since table styling is quite restrictive.

This PR also adds tooltips to any cell that has the truncate method on.

https://amino-git-fix-components-zonos.vercel.app/?path=/story/amino-simpletable--with-link

Here are some screenshots I took when running dashboard while pointing to local amino. I will need to go through each table and make sure they look good in dashboard. 

Screenshot 2024-10-17 at 2 47 48 PM
Screenshot 2024-10-17 at 2 47 31 PM
Screenshot 2024-10-17 at 1 45 41 PM

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.4.9",
+  "version": "5.4.10",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:Zonos/amino.git",
   "license": "MIT",


evad1n:
It may become a reasonable idea for any custom alterations to the table to try and use custom components instead of changing `SimpleTable` directly. Like maybe having a `TableCellTruncated` component that you can pass to `renderCustom` or something. Just trying to come up with ideas to not have this thing grow out of control.

ckgardner:
> > It may become a reasonable idea for any custom alterations to the table to try and use custom components instead of changing `SimpleTable` directly. Like maybe having a `TableCellTruncated` component that you can pass to `renderCustom` or something. Just trying to come up with ideas to not have this thing grow out of control. Especially considering how widely used it is now and how finicky the styling is.
> 
> Yeah... but I also see needing this truncate stuff on like every table. So it's also nice to be able to change with a prop. 😬

Also, really not wanting to refactor rn 😂

goodpickle:
The ellipses goes away when the checkbox is checked. Unsure if that is intentional. 🤷 

Screenshot 2024-10-17 at 3 14 40 PM

Screenshot 2024-10-17 at 3 14 52 PM

ckgardner:
ah my truncating completely broke on a recent push. One sec.

goodpickle (dismissed):

evad1n (approved):

PR #759, opened 2024-11-11
Feat/simple table collapsible

nick-lanners:
## Linear issue
Facilitates this task:
[CXT-1064: Frontend - Add filters to Invoices UI and update side nav bar](https://linear.app/zonos/issue/CXT-1064/frontend-add-filters-to-invoices-ui-and-update-side-nav-bar)

Figma for SimpleTable designs:
<https://www.figma.com/design/sq4q6XA4CjrQWSdnh3vnXM/Billing?node-id=2532-33802&m=dev>

## Preview
Collapsible SimpleTable
<https://simple-table.amino.zonos.com/?path=/story/amino-simpletable--collapsible>

Bordered SimpleTable
<https://simple-table.amino.zonos.com/?path=/story/amino-simpletable--bordered>

## Description

This PR adds new functionality to SimpleTable:
- enable rows to be collapsible/expandable
- enables removal of the headers (for some new designs)
- enables shift of design from "normal" to "bordered" (for some new designs)

## Todo

### DIFF
--- a/src/components/simple-table/SimpleTable.tsx
+++ b/src/components/simple-table/SimpleTable.tsx
@@ -99,7 +99,7 @@ export type SimpleTableProps<T extends object> = BaseProps & {
   bordered?: boolean;
   /**
    * @default false
-   * @note Setting collapsible with selectable will result in unexpected behavior
+   * @note Setting collapsible with onRowClick could result in unexpected behavior
    * Enable rows to collapse and expand with more information
    */
   collapsible?: {
@@ -148,7 +148,6 @@ export type SimpleTableProps<T extends object> = BaseProps & {
   renderFooter?: ReactNode;
   /**
    * @default false
-   * @note Setting collapsible with selectable will result in unexpected behavior
    * Show checkbox on each row, and checkbox for toggling all in header
    */
   selectable?: {
--- a/src/components/simple-table/SimpleTable.tsx
+++ b/src/components/simple-table/SimpleTable.tsx
@@ -1,9 +1,10 @@
-import React, { type ReactNode, Fragment } from 'react';
+import React, { type ReactNode, Fragment, useState } from 'react';
 
 import clsx from 'clsx';
 
 import { Checkbox } from 'src/components/checkbox/Checkbox';
 import { Skeleton } from 'src/components/skeleton/Skeleton';
+import { TableRowCollapse } from 'src/components/table/TableRowCollapse';
 import { Text } from 'src/components/text/Text';
 import { Tooltip } from 'src/components/tooltip/Tooltip';
 import type { BaseProps } from 'src/types/BaseProps';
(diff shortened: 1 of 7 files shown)


src/components/simple-table/SimpleTable.tsx
```diff
@@ -312,6 +346,53 @@ export const SimpleTable = <T extends object>({
...
+                        }
+                        disabled={
+                          selectable.isRowCheckboxDisabled?.(item, index) ||
+                          false
+                        }
+                        onChange={checked =>
+                          selectable.onRowCheckboxChange?.(checked, item, index)
+                        }
+                      />
+                    )}
```
ckgardner, line 379: Is there a storybook way to preview selectable and collapsible?

nick-lanners: Oh it was my intention, at this time to have them be mutually exclusive. Similar with having onRowClick, it'll take some more work to sort out the click listeners between the different "modes". I'll add a couple quick comments about that. Those will come in a later version (probably when a design calls for them).

nick-lanners: Lol jk, selectable actually seems to work pretty well. I'm going to add it to the stories.

evad1n: At this point it may be worth creating a separate table as this one is far from simple now. Originally when I created it I wanted to just be able to pass the items and headers and not have to worry about rendering logic. At that point the way to build collapsible tables was to use custom `TableRowCollapse`.

evad1n: The main advantage in that case being when we make updates specific to collapsing tables or such, that we don't have to worry about breaking other existing tables that are unrelated.

nick-lanners: I don't know. I find the SimpleTable incredibly useful, even as it becomes a bit more "complicated." Being able to just pass some props and the rendering and styles are already taken care of is pretty great, especially after getting used to how the thing works. My hope with these changes was to be able to keep using the code patterns we have throughout Dashboard with the new designs coming out (the bordered table style).

If we want break things apart, I suppose I can start over, but I worry that it will be a bunch of extra code that is essentially the same as SimpleTable for one change in functionality.

Or maybe we just change the name of the component...

ckgardner: hmmm, I see both points of view. I lean towards agreeing with Nick. I don't like the idea of having multiple amino tables that are commonly used and need to be updated separately. But I agree that any change to the simple table imposes a risk on other use cases in dashboard. I wonder if there is a better way to protect against that.

nick-lanners: This might be a conversation for our planning meeting. For what it's worth, everything associated with this change is optional, so it _shouldn't_ change any of the current tables in Dashboard. However, if changes are made to `TableRowCollapse`, it will affect SimpleTables using the `collapsible` functionality.

evad1n (dismissed):

evad1n (approved):

PR #765, opened 2024-11-25
FEAT: Button testIds

ckgardner:
## Description

This PR implements `data-testids` for the `Button` and `MenuButton` components. The input already has this feature. This will help when writing synthetic tests and trying to query select.

## Todo


src/components/button/Button.tsx
```diff
@@ -96,12 +101,27 @@ export function Button<T extends GroupTag = typeof DEFAULT_TAG>({
...
+  }, []);
+
+  const defaultTestId = useMemo(
+    () =>
+      getTestId({
+        componentName: 'button',
+        name: buttonText,
+      }),
+    [buttonText],
+  );
```
Thienhuynh95, line 124: This useCallback hook approach doesn't seem to work, the callback would only trigger once, and when it's triggered, the ref in DOM may not be available to access yet.
image

ckgardner: These are the right names. It's the text within the button appended to `amino--`. Unless you are referencing something else?

Thienhuynh95 (approved):

Thienhuynh95 (approved):

PR #768, opened 2024-11-29
fix: lock collapse button to end of row

nick-lanners:
## Description

This PR makes a small fix to lock the collapse chevron to the end of the row, rather than acting like a normal <td> element

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.4.29",
+  "version": "5.4.30",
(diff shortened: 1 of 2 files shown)


Thienhuynh95 (approved):

evad1n (dismissed):
Do you think we can write a test for this?

nick-lanners:
> Do you think we can write a test for this?

I can try to figure out a a decent test for it. Fight with my Playwright right now, but when I can get it working I'll create one.

evad1n (approved):

PR #771, opened 2024-12-04
Fix/simple table collapse

nick-lanners:
## Description

This PR adjusts some functionality for the SimpleTable, especially around the Collapsible feature:
- handle items with no collapsible content better with CSS
- fix onRowClick functionality with collapsible
- lift collapsed items state out of the component so consumers have more flexibility with how it works
 - ie make network calls or watch state for style changes
- write tests to handle these changes 

## Todo

### DIFF
--- a/src/components/simple-table/__stories__/SimpleTable.stories.tsx
+++ b/src/components/simple-table/__stories__/SimpleTable.stories.tsx
@@ -365,6 +365,13 @@ export const Bordered = () => {
 
 export const Collapsible = () => {
   const [selectedRowIndexes, setSelectedRowIndexes] = useState<number[]>([]);
+  const [expandedItemKeys, setExpandedItemKeys] = useState<string[]>([]);
+  const toggleItem = (id: string) =>
+    setExpandedItemKeys(
+      expandedItemKeys.includes(id)
+        ? expandedItemKeys.filter(x => x !== id)
+        : expandedItemKeys.concat(id),
+    );
 
   const checkboxAllValue = selectedRowIndexes.length === items.length;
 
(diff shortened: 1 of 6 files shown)


src/components/simple-table/__stories__/SimpleTable.stories.tsx
```diff
@@ -618,3 +639,63 @@ export const Custom = () => {
...
+        collapsible={{
+          collapseContent,
+          enabled: true,
+          expandedItemKeys,
+          toggleItem,
+        }}
+        headers={tableHeaders}
+        items={items}
+        keyExtractor={item => String(item.id)}
+        onRowClick={item => alert(`Clicked ${item.name}`)}
```
goodpickle, line 697: Why the alert?

ckgardner:
Screenshot 2024-12-04 at 11 05 09 AM
 The expanded content spacing seems a little off

nick-lanners:
> Screenshot 2024-12-04 at 11 05 09 AM The expanded content spacing seems a little off (almost touching the border)

@ckgardner The content for the collapse is controlled by the consumer, so the css would be controlled on that end. I have designs that don't have any padding and such in the collapse: https://www.figma.com/design/sq4q6XA4CjrQWSdnh3vnXM/Billing?node-id=838-97897&m=dev. I'm not too concerned that the stories look pretty, as long as they show the functionality correctly.

ckgardner:
> > Screenshot 2024-12-04 at 11 05 09 AM The expanded content spacing seems a little off (almost touching the border)
> 
> @ckgardner The content for the collapse is controlled by the consumer, so the css would be controlled on that end. I have designs that don't have any padding and such in the collapse: https://www.figma.com/design/sq4q6XA4CjrQWSdnh3vnXM/Billing?node-id=838-97897&m=dev. I'm not too concerned that the stories look pretty, as long as they show the functionality correctly.

I still think it's worthwhile to have good looking stories to reflect how we want things to look 🤷🏼 , but I get what you are saying with it not being critical.

ckgardner (dismissed):

Thienhuynh95 (approved):

PR #773, opened 2024-12-10
build: update amino

evad1n:
[updates.amino.zonos.com](https://updates.amino.zonos.com/)

## new icons story

https://updates.amino.zonos.com/?path=/story/amino-icons--icons

## flags story

https://updates.amino.zonos.com/?path=/story/amino-icons-flags--flags

## country select

https://updates.amino.zonos.com/?path=/story/amino-select-countryselect--basic-country-select

## Description

This PR:

- fixes caddy 502 interaction with loading too many icons (I couldn't fix this caddy side, so I change how we load icons inside storybook)
- upgrades eslint and friends to latest (v9)
- upgrades storybook to latest
- ~converts to ESM~ was reverted because it's too complicated
- adds storybook tags to stories to show which ones have tests
- remove dead components (kept `ConnectionMap`)
- flag icons now use dynamic imports (react lazy and suspense)
- `Select` now uses intersection observer to only render in view options
- refactor some country selects to work with the new dynamic flag import
- adds a rollup plugin to handle dynamic import file paths. In storybook it needs the `.tsx` extension, but when built we can't have that because it is compiled to JS and needs no extension. I have added a plugin to remove the suffix when building. Tested on dashboard and it works.

tested upstream in dashboard, only breaking changes are the moving of the amino types

## Todo


src/components/rich-radio/RichRadio.tsx
```diff
@@ -63,14 +63,14 @@ export const RichRadio = <T extends string>({
...
         <button
-          key={item.value}
           className={clsx(
             styles.styledItem,
             itemHeight === 40 && styles.smallItem,
           )}
           data-disabled={item.disabled}
           data-state={item.value === selectedValue ? 'checked' : ''}
           data-tip={item.tooltip}
+          key={item.value}
```
Thienhuynh95, line 73: I think we want key to be in front of everything, it looks like the eslint rule is not working right

Thienhuynh95: image

build-utils/css/constants/theme.ts
```diff
@@ -67,81 +66,81 @@ export const theme = {
...
   'line-height-base': '1.5rem',
+  /** @info 24px */
+  'line-height-l': '1.5rem',
   /** @info 16px */
   'line-height-s': '1rem',
+  /** @info 32px */
+  'line-height-xl': '2rem',
   /** @info 12px */
   'line-height-xs': '0.75rem',
+  /** @info 10px */
```
Thienhuynh95, line 123: My idea before for this order was to separate them by group. Ex: `line-height` one section, but it looks like the order is no longer in that order. Also `line-height-2xl` is in `/** OTHER */` group instead of `/** LINE-HEIGH */`. So I think you should at least put the comment tag `/** */` in its right spot.
```
  /** @info 40px */
  'line-height-2xl': '2.5rem',
```

evad1n: Yeah I think I linted these before ignoring them at some point. I will restore the originals

src/components/filter/filter-amount/FilterAmount.tsx
```diff
@@ -8,8 +8,8 @@ import {
 
 import {
   type FilterAmountAction,
-  type FilterAmountState,
   filterAmountOptions,
+  type FilterAmountState,
   initialFilterAmountState,
 } from 'src/components/filter/filter-amount/filterAmountReducer';
```
Thienhuynh95, line 14: I think we discussed this before, just want to make sure this is what we prefer now. `eslint-plugin-simple-import-sort` version > 9 would have a breaking change on the way they organizae the import order, which means `type` and `non-type` will be alphabetically sorted regardless it's type or not.

src/icons/flags/AD.tsx
```diff
@@ -3,9 +3,9 @@ import { forwardRef } from 'react';
 import { FlagIconBase } from 'src/icons/flag-icon/_FlagIconBase';
 
 type Props = {
-  borderRadius?: number;
   height: number;
   width: number;
+  borderRadius?: number;
```
Thienhuynh95, line 8: How come this prop is not alphabetically sorted? is this the new rule where require prop has higher priority? :thinking:

svgReact/icons/config/config.ts
```diff
@@ -9,7 +11,8 @@ export type FileConfig = {
    * @desc Input path that all generated file located
    * @example svgReact/icons/dist
    */
-  inputFolderPath: `svgReact/icons/dist${string}` | `src/icons${string}`;
+  // inputFolderPath: `svgReact/icons/dist${string}` | `src/icons${string}`;
+  inputFolderPath: string;
```
Thienhuynh95, line 15: Do you mean to remove the comment above?

svgReact/icons/index.ts
```diff
@@ -30,7 +30,11 @@ const generateSvgs = ({
     .filter(Boolean)
     /** @desc Preprocessing file */
     .map(item => {
-      const hasSubfolder = item.split('/').length > 1;
+      if (item.split('/').length > 2) {
+        return null;
+      }
```
Thienhuynh95, line 35: Looks like this logic is to not process more than 2 subfolder level deep. Can you add a comment here?

build-utils/css/constants/theme.ts
evad1n: Should be fixed now

src/components/filter/filter-amount/FilterAmount.tsx
evad1n: I think it is fine to use the new default behavior, I don't see it as very important for the type imports to be separate. I'd rather have less custom config.

build-utils/css/constants/theme.ts
Thienhuynh95: Can you also re-build it so the new theme.css will also be updated?

Thienhuynh95 (approved):

Thienhuynh95 (approved):

PR #780, opened 2024-12-20
FIX: Checkbox label

ckgardner:
## Linear issue

[CXT-1344: Frontend - Fix checkbox in Amino](https://linear.app/zonos/issue/CXT-1344/frontend-fix-checkbox-in-amino)

## Description

This PR moves the subtitle out of the checkbox label element. It was causing nested links to be un-clickable and is actually recommended as a poor user experience by [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#interactive_content). Removed unnecessary props because of this change.

Preview: https://cade.amino.zonos.com/?path=/story/amino-checkbox--checkbox-with-complex-subtitle

Country multiselect still works: https://cade.amino.zonos.com/?path=/story/amino-country-multi-select-countrymultiselectexpanded--basic

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.5.2",
+  "version": "5.5.3",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:Zonos/amino.git",
   "license": "MIT",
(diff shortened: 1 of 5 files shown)


goodpickle (approved):

evad1n (approved):

evad1n (changes requested):
https://cade.amino.zonos.com/?path=/story/amino-select-multiselect--basic

evad1n (approved):

PR #791, opened 2025-01-29
Select and input updates

connorvang:
## Description

- Updates the select component to remove the floating labels and instead have optional labels on top of the input
- Updates to the inputs with a new "InputBase" that is the same styling with the label on top of the input.
- Updates other inputs to use the new InputBase instead of the floating input. I kept the floating input there to not be destructive in case there is ever a time where its needed or wanted.

## Todo


ckgardner:
Screenshot 2025-01-30 at 2 58 58 PM

Do we want placeholders inside the input when there is no default value or is this intended?

ckgardner:
Something weird going on with the state of multi select
https://github.com/user-attachments/assets/5cf061dc-58e2-423f-ba20-f7393ef38a40

ckgardner:
The country select won't load properly unless you create a zonos.com domain.
https://amino-git-selectbananza-zonos.vercel.app/?path=/story/amino-select-countryselect--basic-country-select

connorvang:
> Screenshot 2025-01-30 at 2 58 58 PM
> Do we want placeholders inside the input when there is no default value or is this intended?

Yes there would be a placeholder, I just didn't update all of the stories to include one

connorvang:
> Something weird going on with the state of multi select https://github.com/user-attachments/assets/5cf061dc-58e2-423f-ba20-f7393ef38a40

We don't use that component anymore. You could consider it "deprecated" It was only ever used for the countries multi select but Will made a whole new component for it.

connorvang:
> The country select won't load properly unless you create a zonos.com domain. https://amino-git-selectbananza-zonos.vercel.app/?path=/story/amino-select-countryselect--basic-country-select

I was curious why those wouldn't load. I just assumed it was local div thing.

ckgardner:
1. Looks like the "input with error" story needs to be fixed (no red highlight): https://amino-git-selectbananza-zonos.vercel.app/?path=/story/amino-input--input-with-error
2. The prefix has quite a bit of space between icon and text. Intentional?
Screenshot 2025-01-30 at 3 19 54 PM

Everything else looks good in the preview 👍🏼

connorvang:
Yeah I noticed that large space between the icon and text but we don't actually have anywhere in Dashboard that noticed where we include an icon so I left it for a future styling update. Ill get that error updated

src/components/input/input-type/_DateInput.module.scss
```diff
@@ -3,6 +3,15 @@
...
+  input[type='datetime-local']::-webkit-calendar-picker-indicator {
+    display: none;
+    -webkit-appearance: none;
+  }
+
+  input[type='date']::-webkit-calendar-picker-indicator {
+    display: none;
+    -webkit-appearance: none;
+  }
 }
```
ckgardner, line 15:
```suggestion
  input[type='datetime-local']::-webkit-calendar-picker-indicator,  input[type='date']::-webkit-calendar-picker-indicator {
    display: none;
    -webkit-appearance: none;
  }
}
```

ckgardner (changes requested):

ckgardner (dismissed):

ckgardner (approved):

PR #817, opened 2025-06-10
new logos 👀 

connorvang:
## Description

Updates the Zonos logos to be the new logos. Adds the theming support for the logos since they are slightly different between light and dark to avoid an optical illusion on dark backgrounds. 

https://amino-8jmh1fkwq-zonos.vercel.app/?path=/docs/amino-zonos-logos--docs

## Todo


ckgardner (dismissed):
Love it!

src/icons/custom/ZonosIcon.tsx
Thienhuynh95, line 1: This file is being generated from the script, so it's not good to modify it directly. The process is that we have the script that would convert the icon svg to React component. I can show you the process if you want!

Thienhuynh95 (approved):

PR #825, opened 2025-08-06
feat(CountryMultiSelectExpanded): enhance component with new height prop and tooltip support for disabled countries

Thienhuynh95:
### Fixed height table (the height will stay as is when searching)
https://thien.amino.zonos.com/?path=/story/amino-country-multi-select-countrymultiselectexpanded--no-groups-fixed-height
image

### Add the ability to have tooltip on the disabled row
https://thien.amino.zonos.com/?path=/story/amino-country-multi-select-countrymultiselectexpanded--no-groups-with-disabled-tooltip
image

## Description

This PR enhance component with new height prop and tooltip support for disabled countries

## Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zonos/amino",
-  "version": "5.5.35",
+  "version": "5.5.36",
   "description": "Core UI components for Amino",
   "repository": "git@github.com:Zonos/amino.git",
   "license": "MIT",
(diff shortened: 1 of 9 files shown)


ckgardner:
I think you have to create a zonos.com domain for countries to work 🤔

ckgardner (approved):
https://github.com/user-attachments/assets/6e305cba-cadf-4897-99a1-1f7ee6384f6e

This is pre-existing in production and pretty minor, but the height of the input changes probably due to the "ClearIcon" being rendered. It causes the whole thing to shift in height once you type 🤷🏼 You can decide what you wanna do though. ✅

